PR #4, opened 2020-01-10
add utility functions for downloading compressed files

JanekEbb:


paderbox/io/download.py
```diff
@@ -0,0 +1,184 @@
...
+    local_dir = path.dirname(local_file)
+    if not path.isfile(local_file):
+        def progress_hook(t):
+            """
+            Wraps tqdm instance. Don't forget to close() or __exit__()
+            the tqdm instance once you're done with it (easiest using
+            `with` syntax).
+            """
+
+            last_b = [0]
```
boeddeker: How about using `nonlocal` in `inner`? This would be a bit more explicite.

```python
last_b = 0
def inner(...):
  nonlocal last_b
  ...
  last_b = ...
```

```

File: paderbox/io/download.py [hunk 2]
```diff
@@ -0,0 +1,184 @@
...
+
+    Args:
+        remote_file:
+        local_file:
+
+    Returns:
+
+    """
+    local_dir = path.dirname(local_file)
+    if not path.isfile(local_file):
```
boeddeker: How about using an argument `exist_ok` to allow the skip?

paderbox/io/download.py
```diff
@@ -0,0 +1,184 @@
...
+                    remains unchanged.
+                """
+                if tsize is not None:
+                    t.total = tsize
+                t.update((b - last_b[0]) * bsize)
+                last_b[0] = b
+
+            return inner
+
+        tmp_file = path.join(local_dir, 'tmp_file')
```
boeddeker: Maybe change this to a more thread safe name?
e.g. `local_file + '_tmp_file'`

paderbox/io/download.py
```diff
@@ -0,0 +1,184 @@
...
+                                os.remove(
+                                    path.join(local_dir, member.filename)
+                                )
+
+                                # Quit
+                                sys.exit()
+            os.remove(local_file)
+
+        elif local_file.endswith('.tar.gz'):
+            tar = tarfile.open(local_file, "r:gz")
```
boeddeker: Does `tarfile.open` support a `with` context?

paderbox/io/download.py
```diff
@@ -0,0 +1,184 @@
...
+                reporthook=progress_hook(t),
+                data=None
+            )
+        os.rename(tmp_file, local_file)
+    return local_file
+
+
+def extract_file(local_file):
+    """
+    extract files if local_file zip or tar.gz else do nothing
```
boeddeker: Is my understanding correct, that this code skips existing files?

Could you add a description to the doctext for the `# Trick to omit first level folder`?

paderbox/io/download.py
```diff
@@ -0,0 +1,184 @@
...
+                for name in z.namelist():
+                    if not name.endswith('/'):
+                        parts.append(name.split('/')[:-1])
+                prefix = path.commonprefix(parts) or ''
+
+                if prefix:
+                    if len(prefix) > 1:
+                        prefix_ = list()
+                        prefix_.append(prefix[0])
+                        prefix = prefix_
```
boeddeker: What is the difference of this code to
`prefix = [prefix[0]]`?

Or later using:
`prefix = f'/{prefix[0]}/'`

paderbox/io/download.py
```diff
@@ -0,0 +1,184 @@
...
+
+                    prefix = '/'.join(prefix) + '/'
+                offset = len(prefix)
+
+                # Start extraction
+                members = z.infolist()
+                for i, member in enumerate(members):
+                    if len(member.filename) > offset:
```
boeddeker: When is this `if` statement false? What should happen in that case?

paderbox/io/download.py
boeddeker: Could you also add a motivation for the `# Trick to omit first level folder`?

boeddeker:
Is there a reason, why you do not use `pathlib`?

boeddeker:
I discussed with @LukasDrude and all functions in `io` should support `str` and `pathlib.Path` as arguments.
Internally the functions may use `str` or `pathlib.Path`, with a preference for `pathlib.Path`.

paderbox/io/download.py
```diff
@@ -0,0 +1,186 @@
...
+                    if not target_file.exists():
+                        try:
+                            z.extract(member=member, path=local_dir)
+                        except KeyboardInterrupt:
+                            # Delete latest file, since most likely it
+                            # was not extracted fully
+                            if target_file.exists():
+                                os.remove(target_file)
+                            # Quit
+                            sys.exit()
```
boeddeker: Can you change this to `raise`? Or is there a reason to use `sys.exit()`?

paderbox/io/download.py
```diff
@@ -0,0 +1,186 @@
...
+                    if not target_file.exists():
+                        try:
+                            tar.extract(tar_info, local_dir)
+                        except KeyboardInterrupt:
+                            # Delete latest file, since most likely it
+                            # was not extracted fully
+                            if target_file.exists():
+                                os.remove(target_file)
+                            # Quit
+                            sys.exit()
```
boeddeker: See comment for zip file

paderbox/io/download.py
```diff
@@ -0,0 +1,186 @@
...
+        logger:
+
+    Returns:
+
+    """
+
+    target_dir = Path(target_dir)
+    os.makedirs(target_dir, exist_ok=True)
+    # Set socket timeout
+    socket.setdefaulttimeout(120)
```
boeddeker: Why to you set such a timeout?
Does a context manager exist for this function?

boeddeker:
Now I am fine with this PR. I will merge.

PR #5, opened 2020-01-31
create leaner setup.py

jensheit:
removed unnecessary dependencies and allows installation without visualization if needed


boeddeker:
Can you simplify the `setup.py` `extras_require`?
In `pb_bss` we also started with such a complicated `extras_require`.
I discussed this with @LukasDrude and the conclusion was: We have no use case to have different dependencies. Either you want the minimum setup or the maximum.

boeddeker:
In the discussion, what to upload and what to keep internal, we concluded that we upload everything that someone external can use.
`paderbox/utils/strip_solution.py` has nothing to do with our internal environment. Why did you delete that file?

jensheit:
> Can you simplify the `setup.py` `extras_require`?
> In `pb_bss` we also started with such a complicated `extras_require`.
> I discussed this with @LukasDrude and the conclusion was: We have no use case to have different dependencies. Either you want the minimum setup or the maximum.

In my mind just using the visualization tools is a valid way of using the paderbox. The dependency between these tools and the rest of paderbox are minor

JanekEbb:
Why does test refer to test + visualization? I would prefer to only have 'full': visualization + test

jensheit:
okay, I changed the names in the setup to following the last two commend . Do these names match with your expectations? @boeddeker @JanekEbb

jenkins_common.bash
```diff
@@ -28,6 +28,6 @@ source "${TOOLBOX}/bash/kaldi.bash"
 # Refresh toolbox
 pip uninstall --quiet --yes paderbox
 pip show paderbox
-pip install --quiet --user -e ${TOOLBOX}
+pip install --quiet --user -e ${TOOLBOX}[all]
```
boeddeker: In our environment this should be without the `all`. All dependencies should be installed in the global environment.

jensheit:
This pull request is addressed in #11

PR #6, opened 2020-02-21
Audioread Fallback for .m4a files

TCord:
Adds .m4a to supported filetypes of pb.audioread by exchanging SoundFile with audioread for .m4a files.

### DIFF
--- a/paderbox/io/audioread.py
+++ b/paderbox/io/audioread.py
@@ -10,7 +10,7 @@ from pathlib import Path
 
 import numpy as np
 import soundfile
-import audioread
+import audioread as ar
 
 import paderbox.utils.process_caller as pc
 from paderbox.io.path_utils import normalize_path
@@ -168,7 +168,7 @@ def load_audio(
 
     try:
         if path[-3:] == 'm4a':
-            with audioread.audio_open(
+            with ar.audio_open(
                     path
             ) as f:
                 samplerate = f.samplerate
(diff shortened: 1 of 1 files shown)


paderbox/io/audioread.py
```diff
@@ -166,22 +167,35 @@ def load_audio(
...
-                    'PCM_16': np.int16,
-                    'FLOAT': np.float32,
-                    'DOUBLE': np.float64,
-                })
-                dtype = mapping[f.subtype]
-
-            frames = f._prepare_read(start=start, stop=stop, frames=frames)
-            data = f.read(frames=frames, dtype=dtype, fill_value=fill_value)
-        signal, sample_rate = data, f.samplerate
+        if path[-3:] == 'm4a':
```
boeddeker: Could you include the `.`? (i.e. `'.m4a'`)

paderbox/io/audioread.py
```diff
@@ -166,22 +167,35 @@ def load_audio(
...
+            with ar.audio_open(
+                    path
+            ) as f:
```
boeddeker: Can you add an assert that start and stop aren't supported? And add a comment, that audioread does not support them.

boeddeker:
Is it possible to download a file from the internet to make a test that we can read m4a files?

paderbox/io/audioread.py
TCord: The new version now uses suffix() instead of string slicing

TCord:
Unfortunately I could not find any open source .m4a files to create a test

boeddeker:
Ok, I merged this PR

PR #7, opened 2020-02-25
move ArrayIntervall from pb_chime5 to paderbox

boeddeker:
I moved the source code for the ArrayIntervall from pb_chime5 to this repo.
Furthermore, I did some refactoring of the code to get closer to numpy style code.
e.g. `The ArrayIntervall.__init__ ` takes now a numpy array as input and the old init is now `zeros`.


jensheit:
we should probably add a short text at the beginning, explaining why this tool is necessary

jensheit:
should we add this to the array __init__.py?

boeddeker:
ok, I added an assert to the `from_kaldi_segments` and a small description in the `__ini__.py`

PR #8, opened 2020-03-02
Fixing load_audio dtype BytesIO

TCord:
Fixes audioread for data of type BytesIO and adds test case for this error


boeddeker:
Thanks

PR #10, opened 2020-03-03
py_query: support non identifier column names, add globals dict and add return_selection

boeddeker:
Add support for non identifier column names:
 - Old: When a column had a non identifier column name (e.g. whitespace in the name) py_query failed
 - New: The column entry can be accessed with the `kwargs` dict (see doctest)

Add globals dict:
 - The globals dict for the query was not accessible. Now the user can influence the globals dict and `setup_code` is no longer nessesary.
 - example for using a numpy function in the query: `py_query(..., globals={'np': np})` 

Add option return_selection:
 - Allow to return the selection indices instead of `data[return_selection]`

### DIFF
--- a/paderbox/utils/pandas_utils.py
+++ b/paderbox/utils/pandas_utils.py
@@ -67,7 +67,8 @@ def py_query(
         setup_code: legacy argument, Superseded by the globals argument.
             Additional code which runs before the query conditions.
             You may use this for additional imports.
-        globals: Specify some global names. Useful for imports.
+        globals: Specify some global names. Useful for imports. See doctest how
+            to use it.
         return_selection: Whether to return the selection of the data or the
             selection indices.
 
--- a/paderbox/utils/pandas_utils.py
+++ b/paderbox/utils/pandas_utils.py
@@ -12,7 +12,7 @@ def py_query(
         allow_empty_result=False,
         setup_code='',
         globals=None,
-        return_selection=False,
+        return_selected_data=True,
 ):
     """
     Alternative: pd.DataFrame.query:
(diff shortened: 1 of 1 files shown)


paderbox/utils/pandas_utils.py
```diff
@@ -39,17 +42,37 @@ def py_query(
...
              Note: Use index to get access to the index.
         use_pd_query:  Pandas query is much faster but limited
         allow_empty_result:
-        setup_code: Additional code which runs before the query conditions.
+        setup_code: legacy argument, Superseded by the globals argument.
+            Additional code which runs before the query conditions.
             You may use this for additional imports.
+        globals: Specify some global names. Useful for imports.
+        return_selection: Whether to return the selection of the data or the
+            selection indices.
```
jensheit: should this variable be called return_selection_indices?

paderbox/utils/pandas_utils.py
```diff
@@ -39,17 +42,37 @@ def py_query(
...
-            logical and to be a str. For examples see doctest.
+            logical `and` to be a str. For examples see doctest.
              Note: Use index to get access to the index.
         use_pd_query:  Pandas query is much faster but limited
         allow_empty_result:
-        setup_code: Additional code which runs before the query conditions.
+        setup_code: legacy argument, Superseded by the globals argument.
+            Additional code which runs before the query conditions.
             You may use this for additional imports.
+        globals: Specify some global names. Useful for imports.
```
jensheit: it is not clear what globals is doing, could you perhaps add an doctest?

boeddeker: The last example in the doctest is an example for globals. I updated the description, so the user can search in the doctests

paderbox/utils/pandas_utils.py
boeddeker: Hmm. I am not sure. In `pb_bss` selection are the indices.
But here it may be more clear to be more verbose.

michael-kuhlmann: I am also in favor of renaming `return_selection`. Maybe `return_selection_output` might be an option to be consistent with `pb_bss`? But then the logic has to be swapped.

boeddeker: Ok, I changed it to `return_selected_data`.

PR #12, opened 2020-03-19
add async_run_process an async version of run_process

boeddeker:
Does anyone have a better name for this function?

### DIFF
--- a/paderbox/utils/process_caller.py
+++ b/paderbox/utils/process_caller.py
@@ -131,7 +131,7 @@ def run_process(
         ) from e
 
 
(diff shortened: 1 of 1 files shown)


jensheit:
> Does anyone have a better name for this function?

I am fine with the name, a little verbose but that can help.

TCord:
I am also fine with the name. If you want to underline that it is a asynchronous version of run_process, run_process_async might be another alternative

boeddeker:
I renamed the function and since all tests pass, I will merge this PR.

PR #16, opened 2020-06-03
replace librosa fbanks with own implementation supporting frequency w…

JanekEbb:
…arping (e.g. vtlp)

### DIFF
--- a/paderbox/transform/module_fbank.py
+++ b/paderbox/transform/module_fbank.py
@@ -24,6 +24,7 @@ class MelTransform:
             fmin: Optional[int] = 50,
             fmax: Optional[int] = None,
             log: bool = True,
+            eps: float = 1e-18,
     ):
         """
         Transforms linear spectrogram to (log) mel spectrogram.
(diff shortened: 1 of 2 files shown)


jensheit:
Can we at this point remove librosa from our dependencies?

TCord:
> Can we at this point remove librosa from our dependencies?

I think the only place librosa is also used at this moment is in case .m4a files are loaded. If we catch the exception with a  sufficiently explanatory error, removing librosa should be fine.

boeddeker:
The `.m4a` files need `audioread` (dependency of `librosa`). So we could change the dependency `librosa` to `audioread`.

TCord:
I think, the import of audioread already is lazy to have less dependencies. The file format .m4a is, after all, not a very common format. So removing librosa from dependencies without adding a new one should be the best option.

boeddeker:
The dependency `librosa` is not a default dependency, it is included in the `all` dependency.
In the `all` dependency we should list all dependency.

PR #17, opened 2020-06-15
fix load_csv sniffer

boeddeker:
- Use csv.DictReader instead of csv.reader
 - consider the sniffer option and use the dialect
 - add fieldnames argument

### DIFF
--- a/paderbox/io/__init__.py
+++ b/paderbox/io/__init__.py
@@ -69,6 +69,8 @@ __all__ = [
     "update_hdf5",
     "load_pickle",
     "dump_pickle",
+    "load_csv",
+    "loads_csv",
     "mkdir_p",
     "symlink",
     "SummaryEncoder",
--- a/paderbox/io/__init__.py
+++ b/paderbox/io/__init__.py
@@ -36,6 +36,7 @@ from paderbox.io.yaml_module import (
     dump_yaml_unsafe,
     dumps_yaml_unsafe,
 )
+from paderbox.io.csv_module import load_csv, loads_csv
 from paderbox.io.audioread import load_audio, recursive_load_audio
 from paderbox.io.audiowrite import dump_audio, dumps_audio
 from paderbox.io.file_handling import (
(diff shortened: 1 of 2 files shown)


paderbox/io/__init__.py
```diff
@@ -36,6 +36,7 @@
     dump_yaml_unsafe,
     dumps_yaml_unsafe,
 )
+from paderbox.io.csv_module import load_csv, loads_csv
```
thequilo, line 39: Can you add this to `__all__`?

paderbox/io/csv_module.py
```diff
@@ -1,11 +1,35 @@
...
+        dialect:
+            e.g. 'excel', 'excel-tab' and 'unix'.
+            See python documentation for standard library csv for more details.
+        sniffer:
+            Whether to detect the dialect (i.e. delimiter and line separator)
+            When True, ignore dialect.
+        sniffer_sample_length:
+
+    Returns:
+        list of dicts
```
thequilo, line 32: I prefer to have the type annotations in the function definition, e.g.,
```python
def load_csv(
        file,
        *,
        fieldnames: Optional[List[str]]=None,
        dialect: str="excel",
        sniffer: bool=False,
        sniffer_sample_length: int=1024,
) -> List[dict]:
```
But either one is fine

paderbox/io/__init__.py
boeddeker: done

paderbox/io/csv_module.py
boeddeker: done

thequilo:
How can I read a tab-separated file with this function?

boeddeker:
You can use the `sniffer` option. The standard library tries then to detect the `dialect`.

Alternatively you can create a dialect (see https://docs.python.org/3/library/csv.html).

If the speed doesn't matter (i.e. reading 1 file and not 1000 files), I would try the sniffer.

paderbox/io/csv_module.py
```diff
@@ -66,6 +66,9 @@ def load_csv(
     [{'a': '1', 'b': '2', 'c': '3'}, {'a': '4', 'b': '5', 'c': '6'}]
     >>> loads_csv(content, fieldnames=['d', 'e', 'f'], sniffer=True)
     [{'d': 'a', 'e': 'b', 'f': 'c'}, {'d': '1', 'e': '2', 'f': '3'}, {'d': '4', 'e': '5', 'f': '6'}]
+    >>> content = 'a\\tb\\tc\\n1\\t2\\t3\\n4\\t5\\t6'
```
thequilo, line 69: Are those double back-slashes correct?

boeddeker: Yes, they are correct. The docstring is interpreted from python and you have to escape some characters.
I tested if I can use `r'...'`. It worked, so I removed the double back-slash,

PR #18, opened 2020-08-17
Move get_new_subdir from padertorch to paderbox

boeddeker:
Changes:
 - Rename get_new_folder to get_new_subdir
 - Add prefix and suffix and remove try_id
 - Add cli
 - Add id_naming option

The first three commits are minor changes.

The discussion for this functions started in https://github.com/fgnt/padertorch/pull/46

### DIFF
--- a/paderbox/io/audioread.py
+++ b/paderbox/io/audioread.py
@@ -5,6 +5,7 @@ import inspect
 import os
 import tempfile
 import wave
+import types
 from io import BytesIO
 from pathlib import Path
 import functools
@@ -276,7 +277,7 @@ def recursive_load_audio(
     kwargs = locals().copy()
     path = kwargs.pop('path')
 
-    if isinstance(path, (tuple, list)):
+    if isinstance(path, (tuple, list, types.GeneratorType)):
         data = [recursive_load_audio(a, **kwargs) for a in path]
 
         np_data = np.array(data)
--- a/paderbox/io/new_subdir.py
+++ b/paderbox/io/new_subdir.py
@@ -71,7 +71,7 @@ def get_new_subdir(
         consider_mpi:
             If True, only search on one mpi process for the folder and
             distribute the folder name.
-            When using mpi with this option disables the following can/will
+            When using mpi (and `consider_mpi is False`) the following can/will
             happen
              - When mkdir is True every process will get another folder.
                i.e. each process has a folder just for this process.
--- a/paderbox/io/new_subdir.py
+++ b/paderbox/io/new_subdir.py
@@ -27,16 +27,16 @@ def _removesuffix(self: str, suffix: str) -> str:
 
 
 def get_new_subdir(
-        basedir,
+        basedir: [str, Path],
         *,
-        id_naming='index',
-        mkdir=True,
-        prefix=None,
-        suffix=None,
-        consider_mpi=False,
-        dry_run=False,
+        id_naming: [str, callable]='index',
+        mkdir: bool=True,
+        prefix: str=None,
+        suffix: str=None,
+        consider_mpi: bool=False,
+        dry_run: bool=False,
 ):
-    """Determine a new non existant sub directory.
+    """Determine a new non-existent sub directory.
 
     Features:
      - With mkdir: Thread and process save.
@@ -45,7 +45,7 @@ def get_new_subdir(
 
     Args:
         basedir:
-            The new folder will be inside this directory
+            The new subdir will be inside this directory
         id_naming:
             The id naming that is used for the folder name.
              - str: 'index':
@@ -75,12 +75,13 @@ def get_new_subdir(
             happen
              - When mkdir is True every process will get another folder.
                i.e. each process has a folder just for this process.
-             - When mkdir is False it is likely that all get the same folder,
-               but this is not guarantied. Never use mpi with consider_mpi and
-               mkdir disabled.
+             - Warning: Never use mpi, when `mkdir is False` and
+               `consider_mpi is False`. Depending on some random factors
+               (e.g. python startup time) all workers could get the same
+               folder, but mostly some get the same folder and some different.
+               You never want this.
         dry_run:
-            Similar to mkdir. When true, disable mkdir and print the folder
-            name.
+            When true, disables mkdir and prints the folder name.
 
     Returns:
         pathlib.Path of the new subdir
--- a/paderbox/io/new_subdir.py
+++ b/paderbox/io/new_subdir.py
@@ -1,0 +1,341 @@
+import os
+import re
+import time
+import fnmatch
+import datetime
+import urllib.request
+from pathlib import Path
+
+from scipy._lib.decorator import getfullargspec
+
+
+def _removeprefix(self: str, prefix: str) -> str:
+    # Backport from Python 3.9: https://www.python.org/dev/peps/pep-0616/
+    if self.startswith(prefix):
+        return self[len(prefix):]
+    else:
+        return self[:]
+
+
+def _removesuffix(self: str, suffix: str) -> str:
+    # Backport from Python 3.9:  https://www.python.org/dev/peps/pep-0616/
+    # suffix='' should not call self[:-0].
+    if suffix and self.endswith(suffix):
+        return self[:-len(suffix)]
+    else:
+        return self[:]
+
+
+def get_new_subdir(
+        basedir,
+        *,
+        id_naming='index',
+        mkdir=True,
+        prefix=None,
+        suffix=None,
+        consider_mpi=False,
+        dry_run=False,
+):
+    """Determine a new non existant sub directory.
+
+    Features:
+     - With mkdir: Thread and process save.
+     - Different conventions for ID naming possible, default running index.
+     - MPI aware: Get the folder on one worker and distribute to others.
+
+    Args:
+        basedir:
+            The new folder will be inside this directory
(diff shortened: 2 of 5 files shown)


paderbox/io/audioread.py
```diff
@@ -276,7 +277,7 @@ def recursive_load_audio(
     kwargs = locals().copy()
     path = kwargs.pop('path')
 
-    if isinstance(path, (tuple, list)):
+    if isinstance(path, (tuple, list, types.GeneratorType)):
```
thequilo, line 280: Is this related to `get_new_subdir`?

paderbox/io/new_subdir.py
```diff
@@ -0,0 +1,341 @@
...
+        basedir,
+        *,
+        id_naming='index',
+        mkdir=True,
+        prefix=None,
+        suffix=None,
+        consider_mpi=False,
+        dry_run=False,
+):
+    """Determine a new non existant sub directory.
```
thequilo, line 39: Typo: existent (maybe: non-existent)

paderbox/io/new_subdir.py
```diff
@@ -0,0 +1,341 @@
...
+
+def get_new_subdir(
+        basedir,
+        *,
+        id_naming='index',
+        mkdir=True,
+        prefix=None,
+        suffix=None,
+        consider_mpi=False,
+        dry_run=False,
```
thequilo, line 37: Add type annotations?

paderbox/io/new_subdir.py
```diff
@@ -0,0 +1,341 @@
...
+                This option ensures, that only one program gets the '2' and the
+                remaining programs search for another free id.
+        prefix:
+            Optional prefix for the id. e.g.: '2' -> '{prefix}_2'
+        suffix:
+            Optional suffix for the id. e.g.: '2' -> '2_{suffix}'
+        consider_mpi:
+            If True, only search on one mpi process for the folder and
+            distribute the folder name.
+            When using mpi with this option disables the following can/will
```
thequilo, line 74: disabled? If disables is correct, I don't get this sentence.

paderbox/io/new_subdir.py
```diff
@@ -0,0 +1,341 @@
...
+        if dry_run:
+            print(f'dry_run: "os.makedirs({basedir})"')
+            # ToDo: Make this working.
+            #       Will fail when calling os.listdir
+        else:
+            basedir.mkdir(parents=True)
+
+    if Path('/net') in basedir.parents:
+        # If nt filesystem, assert not in /net/home
+        assert Path('/net/home') not in basedir.parents, basedir
```
thequilo, line 120: Do we want to have this hard-coded in here or flexible as a parameter for easier use in other environments?

paderbox/io/new_subdir.py
```diff
@@ -0,0 +1,341 @@
...
+                dir_nrs += [0]
+                _id = max(dir_nrs) + 1
+                _id = f'{prefix_}{_id}{_suffix}'
+        elif id_naming == 'time':
+            if i != 0:
+                time.sleep(1)
+            _id = datetime.datetime.now().strftime('%Y-%m-%d-%H-%M-%S')
+            _id = f'{prefix_}{_id}{_suffix}'
+        elif callable(id_naming):
+            _id = id_naming()
```
thequilo, line 157: Check if the ID is unique in the subfolder? This is very unlikely for our naming functions below, but this might not be the case for a user-supplied naming function.

paderbox/io/new_subdir.py
```diff
@@ -0,0 +1,341 @@
...
+            suffix:
+            consider_mpi:
+            dry_run:
+
+        Returns:
+
+        """
+        # Add more cli mappings if necessary
+        if id_naming == 'adjective_color_animal':
+            id_naming = NameGenerator(('adjectives', 'colors', 'animals'))
```
thequilo, line 337: Put this into `get_new_subdir`?

paderbox/io/audioread.py
boeddeker: No, I committed this some days ago and forgot to push.
Since this is a minor change, I kept it in this PR.

paderbox/io/new_subdir.py
```diff
@@ -0,0 +1,341 @@
...
+    """Determine a new non existant sub directory.
+
+    Features:
+     - With mkdir: Thread and process save.
+     - Different conventions for ID naming possible, default running index.
+     - MPI aware: Get the folder on one worker and distribute to others.
+
+    Args:
+        basedir:
+            The new folder will be inside this directory
```
jensheit, line 48: The new subdir will be inside ..

paderbox/io/new_subdir.py
```diff
@@ -0,0 +1,341 @@
...
+            distribute the folder name.
+            When using mpi with this option disables the following can/will
+            happen
+             - When mkdir is True every process will get another folder.
+               i.e. each process has a folder just for this process.
+             - When mkdir is False it is likely that all get the same folder,
+               but this is not guarantied. Never use mpi with consider_mpi and
+               mkdir disabled.
+        dry_run:
+            Similar to mkdir. When true, disable mkdir and print the folder
```
jensheit, line 82: Why is this similar to mkdir?

Did you mean:
When true, disables mkdir and prints the folder name
?

paderbox/io/new_subdir.py
```diff
@@ -0,0 +1,341 @@
...
+            basedir.mkdir(parents=True)
+
+    if Path('/net') in basedir.parents:
+        # If nt filesystem, assert not in /net/home
+        assert Path('/net/home') not in basedir.parents, basedir
+
+    prefix_ = f'{prefix}_' if prefix else ''
+    _suffix = f'_{suffix}' if suffix else ''
+
+    for i in range(200):
```
jensheit, line 125: The 200 are a bit arbitrary, should we set a variable somewhere so that it can be changed more easily?

paderbox/io/new_subdir.py
boeddeker: I am also not a fan for hard-coded things, but this one in special:
 - We recommend our students to code in the home directory, while storing the experiment files on a different file system.
    - We do this, because the home directory has an expensive backup system that is useful for code (i.e. never lose your written code).
    - Many other frameworks (e.g. Kaldi, ESPnet, Asteroid, ...) have the code and experiment files in the same folder, so I expect this split is special for us.
      - So others often don't care about this split.
- The naming of `/net/home` is very special. I don't expect that anyone will get a problem with this line. 
   - When someone reports, that a name collision, I will erase this assert.

paderbox/io/new_subdir.py
boeddeker: I don't get your point. The code later will check if the folder exist.
If it is the case, it will call again `id_naming` to get another name.
This is done up to 100 times, before the code complains, that it cannot find a new name.

paderbox/io/new_subdir.py
boeddeker: I am not sure, if I should put this inside (Actually I was unsure if the cli should have this).
On the python side using `NameGenerator(('adjectives', 'colors', 'animals'))` is more verbose and it is obvious, hot to change the generation.

paderbox/io/new_subdir.py
thequilo: Ups, I didn't get that. Then it's fine!

paderbox/io/new_subdir.py
boeddeker: It is even more arbitrary, because later the code ends the loop with the check `i > 100`.
I don't think we have to expose this parameter.

When someone gets a problem with this number, we can change it.
From my observations, it fails fast enough, when something is wrong and launching 100 experiments at the same time is difficult (When you launch with a delay of 10 ms it is unlikely that you get more than 1 collision).

paderbox/io/new_subdir.py
boeddeker: Done

paderbox/io/new_subdir.py
boeddeker: Done.

paderbox/io/new_subdir.py
boeddeker: I changed the complete statement.

paderbox/io/new_subdir.py
boeddeker: done

paderbox/io/new_subdir.py
boeddeker: > Why is this similar to mkdir?

Deleted

>  Did you mean:
> When true, disables mkdir and prints the folder name
> ?

Done

paderbox/io/new_subdir.py
boeddeker: Just for documentation: The origin for this loop is https://github.com/IDSIA/sacred/pull/285.

paderbox/io/new_subdir.py
thequilo: okay

paderbox/io/new_subdir.py
thequilo: I still don't get this sentence

boeddeker: Sorry, I changed the text below. Now I updated this line to
`When using mpi (and `consider_mpi is False`) the following can/will happen`

PR #25, opened 2020-10-22
Bugfixes for nested_merge

thequilo:
- Fix shortcut for mappings with additional arguments
 - Assert mixed updates with scalars and mappings

### DIFF
--- a/paderbox/utils/nested.py
+++ b/paderbox/utils/nested.py
@@ -154,13 +154,24 @@ def nested_merge(default_dict, *update_dicts, allow_update=True, inplace=False):
     AssertionError: [{'bar': 1}, 0]
     >>> nested_merge({'foo': {'bar': 1}}, {'blub': 0}, allow_update=False)
     {'foo': {'bar': 1}, 'blub': 0}
+    >>> nested_merge({'foo': {'blub': 0}}, {'foo': 1}, {'foo': {'bar': 1}})
+    {'foo': {'bar': 1}}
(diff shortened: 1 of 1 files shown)


paderbox/utils/nested.py
```diff
@@ -160,7 +160,7 @@ def nested_merge(default_dict, *update_dicts, allow_update=True, inplace=False):
         if inplace:
             return default_dict
         else:
-            return default_dict.__class__(default_dict)
+            return copy.deepcopy(default_dict)
```
boeddeker, line 163: Why do you change to `deepcopy`? `copy` does exactly what is needed here:
 - Only copy the outer dict, not the values (i.e. avoid a copy of the "leafs")
 - The recursive call takes care to handle the inner dicts.

thequilo: The non-shortcut code path implicitly performs a deep copy, so the shortcut (no update) code path should do the same. There is no recursive call here.

thequilo: Ok, it only performs a deep copy for the mapping types and not for any other types

boeddeker: Oh, I was not carefully checking the code. Here is nothing recursive.
Nonetheless, we should not copy the leafs.
The leafs could be large and then the copy is slow and increases the memory consumption.

I would keep the old behavior to have a trade off between speed and what the user expects.

jensheit (approved):

boeddeker (approved):

PR #28, opened 2020-11-25
Add function for consistent random_state

jensheit:


paderbox/utils/random_utils.py
```diff
@@ -2,6 +2,18 @@
...
 
+def str_to_random_state(string):
+    """
+    This functions outputs a consistent random state dependent on
+    an input string.
+    """
+    from hashlib import md5
+    hash_value = md5(string.encode())
+    hash_value = int(hash_value.hexdigest(), 16)
+    hash_value = hash_value % 2 ** 32 - 1
```
boeddeker, line 13: I would say the `-1` is wrong:
```python
np.random.seed(-1)  # fails
np.random.seed(0)  # works
np.random.seed(2**32 - 1)  # works
np.random.seed(2**32)  # fails

np.random.seed(0 % 2**32)  # works
np.random.seed(2**32 % 2**32)  # works
```

paderbox/utils/random_utils.py
```diff
@@ -2,6 +2,18 @@
 from functools import wraps
 
 
+def str_to_random_state(string):
+    """
+    This functions outputs a consistent random state dependent on
+    an input string.
+    """
+    from hashlib import md5
+    hash_value = md5(string.encode())
```
boeddeker, line 11: In https://github.com/fgnt/sms_wsj/blob/f0cc4402a7111cd603c617a21170b0e0ddb90d48/sms_wsj/database/create_rirs.py#L63 we use sha256, can you use that?
Also, `encode("utf-8")` should be more robust.

paderbox/utils/random_utils.py
jensheit: agreed

paderbox/utils/random_utils.py
```diff
@@ -2,6 +2,18 @@
 from functools import wraps
 
 
+def str_to_random_state(string):
+    """
+    This functions outputs a consistent random state dependent on
+    an input string.
```
boeddeker, line 8: Could you add a tiny doctest? 
`print(str_to_random_state('Some text'))`

boeddeker (approved):

PR #29, opened 2020-11-26
updates io.hdf5 after changes in h5py in update to version 3.0.0

jensheit:


paderbox/io/hdf5.py
```diff
@@ -358,18 +360,37 @@ def ckeck_exists():
...
-                                 str, complex, int, float)):
+                                 complex, int, float)):
                 ckeck_exists()
 
                 h5file[cur_path] = item
 
                 # NaN compares negative in Python.
                 # dataset.value has been deprecated. Use dataset[()] instead.
                 if not h5file[cur_path][()] == item and \
+                        not isinstance(item, str) and \
```
boeddeker, line 371: Co you remove this line? It is no longer necessary

paderbox/io/hdf5.py
```diff
@@ -358,18 +360,37 @@ def ckeck_exists():
...
+                h5file[cur_path] = np.void(item)
+                if not h5file[cur_path][()] == np.void(item):
+                    raise ValueError('The data representation in the HDF5 '
+                                     'file does not match the original dict.')
+            elif isinstance(item, str):
+                ckeck_exists()
+                h5file[cur_path] = item
+
+                if LooseVersion(h5py.__version__) >= '3.0.0':
+                    test_item = item.encode('utf-8')
```
boeddeker, line 388: Maybe we should add here a comment, that h5py changes in version 3 all strings to bytes and cannot distinguish between str and bytes. For bytes, we use the suggestion from the h5py documentation, how raw bytes streams can be stored.

paderbox/io/hdf5.py
```diff
@@ -467,6 +488,10 @@ def __recursively_load_dict_contents_from_group__(cls, h5file, path):
                 if isinstance(ans[key], str):
                     if ans[key] == 'None':
                         ans[key] = None
+                if isinstance(ans[key], bytes):
```
boeddeker, line 491: Can you do first the bytes check and then the `None` check?

boeddeker (approved):

PR #32, opened 2020-12-02
Pass through keep_type in nested_op

thequilo:


paderbox/utils/nested.py
```diff
@@ -281,6 +281,7 @@ def nested_op(
                 broadcast=broadcast,
                 mapping_type=mapping_type,
                 sequence_type=sequence_type,
+                keep_type=keep_type
```
boeddeker, line 284: Can you add a final `,`?

thequilo: Done

boeddeker (approved):

PR #33, opened 2020-12-07
Rename array interval

thequilo:


paderbox/array/interval/core.py
```diff
@@ -166,32 +163,38 @@ class ArrayInterval:
...
+        way.
+
         Args:
             array: 
             inverse_mode:
                 Flag to indicate what the intervals represent:
-                 - False: Intervals represent True
-                 - True: Intervals represent False
+                 - False: Intervals represent activity (i.e., `True`)
+                 - True: Intervals represent silcence (i.e., `False`)
```
boeddeker, line 179: The `inverse_mode` flag is an internal detail and is necessary to implement `np.ones` with an unknown length.
It has nothing to do with `activity` or `silence`.

boeddeker (approved):

PR #34, opened 2020-12-10
Updates in random_utils: sample scalars when no shape stated, add dty…

JanekEbb:
…pe decorator, update dtype defaults, add more sampling functions/clsases

### DIFF
--- a/paderbox/utils/random_utils.py
+++ b/paderbox/utils/random_utils.py
@@ -1,5 +1,7 @@
 import numpy as np
 from functools import wraps
+from scipy.stats import truncnorm, truncexpon
+import dataclasses
 
 
 def str_to_random_state(string):
@@ -25,9 +27,8 @@ def _force_correct_shape(f):
     """
     @wraps(f)
     def wrapper(*shape, **kwargs):
-        if not shape:
-            shape = (1,)
-        elif isinstance(shape[0], (tuple, list)):
+        if len(shape) > 0 and isinstance(shape[0], (tuple, list)):
+            assert len(shape) == 1, shape
             shape = shape[0]
 
         return f(*shape, **kwargs)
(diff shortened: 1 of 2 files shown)


paderbox/utils/random_utils.py
```diff
@@ -25,73 +27,290 @@ def _force_correct_shape(f):
     """
     @wraps(f)
     def wrapper(*shape, **kwargs):
-        if not shape:
-            shape = (1,)
-        elif isinstance(shape[0], (tuple, list)):
+        if len(shape) > 0 and isinstance(shape[0], (tuple, list)):
```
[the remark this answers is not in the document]
thequilo: And the docstring doesn't seem to fit the function. I don't see any seed or SNR-related stuff in the function

paderbox/utils/random_utils.py
```diff
@@ -25,73 +27,290 @@ def _force_correct_shape(f):
...
+    return truncexpon(truncation / scale, loc, scale).rvs(shape)
+
+
+@_force_correct_shape
+def hermitian(*shape, dtype=np.complex128):
     """ Assures a random positive-semidefinite hermitian matrix.
 
     :param shape:
-    :param data_type:
+    :param dtype:
```
thequilo, line 240: Same as above (and for the following function): Can make sure that all docstrings use the same style?

paderbox/utils/random_utils.py
```diff
@@ -25,73 +27,290 @@ def _force_correct_shape(f):
...
     :return:
     """
-    matrix = hermitian(*shape, data_type=data_type)
+    matrix = hermitian(*shape, dtype=dtype)
     matrix += np.broadcast_to(shape[-1] * 2 * np.eye(shape[-1]), shape)
     return matrix
+
+
+@dataclasses.dataclass
+class Uniform:
```
thequilo, line 266: I'd add a `dtype` attribute to all of these classes to match the function interfaces

paderbox/utils/random_utils.py
```diff
@@ -25,73 +27,290 @@ def _force_correct_shape(f):
...
+                return np.array(f(*shape, **kwargs), dtype=dtype_local)
+
+            if dtype in (np.float32, np.float64):
+                return _f(dtype)
+            elif dtype is np.complex64:
+                return _f(np.float32) + 1j * _f(np.float32)
+            elif dtype is np.complex128:
+                return _f(np.float64) + 1j * _f(np.float64)
+            else:
+                raise ValueError(f'Invalid dtype {dtype}')
```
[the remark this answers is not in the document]
JanekEbb: Given that all sampling functions in this file provide continuous values I don't think integer types should be supported here.

paderbox/utils/random_utils.py
```diff
@@ -18,80 +20,325 @@ def str_to_random_state(string):
...
 
-    :param f: Function to be wrapped
-    :return: noise_signal
     """
     @wraps(f)
     def wrapper(*shape, **kwargs):
-        if not shape:
-            shape = (1,)
-        elif isinstance(shape[0], (tuple, list)):
+        if len(shape) > 0 and isinstance(shape[0], (tuple, list)):
```
boeddeker, line 33: Does this fail, when `shape is None`?

paderbox/utils/random_utils.py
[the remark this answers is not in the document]
boeddeker: I agree. For the functions here, integer is not a good choice

paderbox/utils/random_utils.py
```diff
@@ -18,80 +20,325 @@ def str_to_random_state(string):
...
+    >>> x.dtype
+    dtype('float64')
+    >>> x = uniform(2, 3)
+    >>> x.shape, x.dtype
+    ((2, 3), dtype('float64'))
+    >>> x = uniform(2, 3, dtype=np.complex128)
+    >>> x.shape, x.dtype
+    ((2, 3), dtype('complex128'))
+    """
+    return np.random.uniform(low=low, high=high, size=None if len(shape) == 0 else shape)
```
boeddeker, line 97: I thought `size=shape` should work when `len(shape) == 0`.

paderbox/utils/random_utils.py
```diff
@@ -1,5 +1,7 @@
 import numpy as np
 from functools import wraps
+from scipy.stats import truncnorm, truncexpon
+import dataclasses
```
boeddeker, line 6: can you add an `__all__` dict?

paderbox/utils/random_utils.py
```diff
@@ -18,80 +20,325 @@ def str_to_random_state(string):
...
+    """
+    return scale * randn(*shape, dtype=dtype) + loc
 
 
 @_force_correct_shape
-def hermitian(*shape, data_type=np.complex128):
+@_add_kwarg_dtype(default=np.float64)
+def truncated_normal(*shape, loc=0., scale=1., truncation=3.):
+    """
+
```
boeddeker, line 169: Can you add a small description, what truncation means?

paderbox/utils/random_utils.py
JanekEbb: shape cannot be None here, can it? or do you mean when shape=(None,) ?

paderbox/utils/random_utils.py
JanekEbb: yes you are right

paderbox/utils/random_utils.py
```diff
@@ -37,59 +37,90 @@ def str_to_random_state(string):
...
+class _Sampler:
+    dtype: type = np.float64
 
-    """
-    @wraps(f)
-    def wrapper(*shape, **kwargs):
-        if len(shape) > 0 and isinstance(shape[0], (tuple, list)):
-            assert len(shape) == 1, shape
-            shape = shape[0]
+    def _sample(self, *shape):
```
thequilo, line 51: Why the asterisk expansion? This is a private method and it would work just fine if shape would be passed directly.

thequilo (approved):

paderbox/utils/random_utils.py
boeddeker: Ah, you are right, my fault. The code is correct.

boeddeker (approved):

PR #35, opened 2020-12-11
migrate warping functionality (VTLP) from padertorch.contrib to paderbox

JanekEbb:


paderbox/transform/module_fbank.py
```diff
@@ -73,7 +88,30 @@ def ifbanks(self):
         return np.linalg.pinv(self.fbanks.T).T
 
     def __call__(self, x):
-        x = np.dot(x, self.fbanks)
+        if self.warping_fn is None:
+            x = x @ self.fbanks
+        else:
+            independent_axis = [ax if ax >= 0 else x.ndim+ax for ax in self.independent_axis]
```
boeddeker, line 94: Is `[ax % x.ndim for ax in self.independent_axis]` more obvious?

paderbox/transform/module_fbank.py
```diff
@@ -73,7 +88,30 @@ def ifbanks(self):
...
+            fbanks = get_fbanks(
+                n_mels=self.n_mels,
+                fft_length=self.fft_length,
+                sample_rate=self.sample_rate,
+                fmin=self.fmin,
+                fmax=self.fmax,
+                warping_fn=self.warping_fn,
+                size=size,
+            ).astype(np.float32)
+            fbanks = fbanks / (fbanks.sum(axis=-1, keepdims=True) + 1e-6)
```
boeddeker, line 109: Use `self.eps`?

paderbox/transform/module_fbank.py
```diff
@@ -73,7 +88,30 @@ def ifbanks(self):
...
+            fbanks = fbanks.swapaxes(-2, -1)
+            if fbanks.shape[-3] == 1:
+                x = x @ fbanks.squeeze(-3)
```
boeddeker, line 112: Why is a squeeze necessary? Especially on a different axis, depending on if one dimension is a singleton?

paderbox/transform/module_fbank.py
thequilo: What if `ax < -x.ndim`?

paderbox/transform/module_fbank.py
JanekEbb: Given an x of, e.g., shape (B,T,F) one of the following matrix multiplications may be performed
- (B,T,1,F) @ (B,T,F,N) when fbanks have shape (B,T,F,N)
- (B,T,F) @ (B,F,N) when fbanks have shape (B,1,F,N)

boeddeker: Ah, now I have an idea, where the problem is. Does `np.einsum('...F,...FN->...N')` do the correct operation in all cases?

JanekEbb: It does the correct operations but when I timed it, it was much slower than the current approach.

I run the following timing:

```
import numpy as np

def matmul1(a, b):
    if b.shape[-3] == 1:
        a = a @ b.squeeze(-3)
    else:
        a = (a[..., None, :] @ b).squeeze(-2)
    return a

def matmul2(a, b):
    a = (a[..., None, :] @ b).squeeze(-2)

def matmul3(a, b):
    return np.einsum('...F,...FN->...N',a,b)

a = np.random.randn(32,1000,257)
b = np.random.randn(32,1,257,80)

%timeit matmul1(a,b)
%timeit matmul2(a,b)
%timeit matmul3(a,b)
```
and got

32.8 ms ± 29.5 µs per loop (mean ± std. dev. of 7 runs, 10 loops each),
93.6 ms ± 12.9 µs per loop (mean ± std. dev. of 7 runs, 10 loops each),
566 ms ± 152 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)

for matmul1, matmul2 and matmul3, respectively.

boeddeker: Ah, ok, that is a big difference.
I didn't expect such a difference, I never profiled dot vs batched dot vs einsum loop.

How about a comment like the following in the source code?
```
# The following ts the same as `np.einsum('...F,...FN->...N', x, fbanks)`,
but much faster (see https://github.com/fgnt/paderbox/pull/35)
```
So no one will simplify the code in the future to the einsum code.

paderbox/transform/module_fbank.py
```diff
@@ -20,31 +20,44 @@ def __init__(
...
         Args:
+            n_mels: number of filters to be applied
             sample_rate: sample rate of audio signal
             fft_length: fft_length used in stft
-            n_mels: number of filters to be applied
             fmin: lowest frequency (onset of first filter)
             fmax: highest frequency (offset of last filter)
             log: apply log to mel spectrogram
+            eps:
+            warping_fn: function to (randomly) remap fbank center frequencies
```
michael-kuhlmann, line 43: Maybe you could also add a sentence about `independent_axis`here, e.g., something similar to the description of `size` in `get_fbanks`.

michael-kuhlmann (approved):

paderbox/transform/module_fbank.py
```diff
@@ -247,6 +248,196 @@ def hz2bin(f, sample_rate, fft_length):
     return f * fft_length / sample_rate
 
 
+def hz_warping(f, alpha, fhi_ratio, fmax=None):
```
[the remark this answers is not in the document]
JanekEbb: not really, alpha and fhi are the variables used in the VTLP paper, therefore probably the most descriptive ones.

paderbox/transform/module_fbank.py
```diff
@@ -247,6 +248,196 @@ def hz2bin(f, sample_rate, fft_length):
...
+    """performs piece wise linear warping of frequencies [Hz].
```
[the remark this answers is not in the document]
boeddeker: I think, he wanted to say, that the notation `frequencies in Hz` is better than `frequencies [Hz]`

boeddeker:
What is the difference between `max_frequency` and `highest_frequency`?

JanekEbb:
highest_frequency is the offset of the last filter in the filterbank, while max_frequency is the maximal frequency in the warping curve. I am also not very happy with the names to be honest. Do you have better suggestions?

boeddeker:
> highest_frequency is the offset of the last filter in the filterbank, while max_frequency is the maximal frequency in the warping curve. I am also not very happy with the names to be honest. Do you have better suggestions?
> 
> Edit: I renamed max_frequency to highest_frequency as well to suggest that these values should be the same

I have no better names. Using the same name for both is probably better for a user, since they mean the same.

paderbox/transform/module_fbank.py
```diff
@@ -85,166 +135,342 @@ def inverse(self, x):
...
+    )
+    f_warped = warp_factor[..., None] * frequency
+    idx = frequency > breakpoints
+    f_warped[idx] = (
+            highest_frequency
+            + (
+                    (frequency[idx] - highest_frequency)
+                    * (highest_frequency - bp_value[idx]) / (highest_frequency - breakpoints[idx])
+        )
+    )
```
michael-kuhlmann, line 340: If `boundary_frequency_ratio` happens to be greater as 1 (e.g., wrong sampling function), this part throws a divide-by-zero error:
```python
paderbox/paderbox/transform/module_fbank.py:338: RuntimeWarning: divide by zero encountered in true_divide
  * (highest_frequency - bp_value[idx]) / (highest_frequency - breakpoints[idx])
```
Since the docstring says that `boundary_frequency_ratio` must be in [0, 1], maybe you could add an assert for that.

thequilo (approved):

boeddeker (approved):

paderbox/transform/module_fbank.py
```diff
@@ -72,185 +98,462 @@ def ifbanks(self):
...
-
-    return feature
-
-
 def get_fbanks(
-        n_mels, sample_rate, fft_length, fmin=0., fmax=None, warping_fn=None
+        sample_rate: int, stft_size: int, number_of_filters: int,
+        lowest_frequency: float = 0.,
+        highest_frequency: Optional[float] = None,
+        warping_fn: Optional[callable] = None,
```
thequilo, line 142: I think it is better to use `typing.Callable` here

thequilo (approved):

thequilo (approved):

PR #37, opened 2021-01-21
Add str_to_seed and str_to_random_generator

thequilo:
Adds `str_to_random_generator` that creates a new `np.random.Generator` and makes seed size configurable.

### DIFF
--- a/paderbox/utils/random_utils.py
+++ b/paderbox/utils/random_utils.py
@@ -23,19 +23,47 @@ __all__ = [
 ]
 
 
-def str_to_random_state(string):
+def str_to_seed(string: str, bits: int = 32) -> int:
     """
-    This functions outputs a consistent random state dependent on
-    an input string.
-    
-    >>> print(str_to_random_state(''))
-    RandomState(MT19937)
+    This functions outputs a consistent seed in the range (0, 2**bits - 1)
+    dependent on an input string.
+
+    >>> print(str_to_seed(''))
+    2018687061
     """
     import hashlib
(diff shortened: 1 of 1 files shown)


paderbox/utils/random_utils.py
```diff
@@ -23,19 +23,47 @@
...
+    >>> print(str_to_random_state(''))
+    RandomState(MT19937)
+    """
+    return np.random.RandomState(str_to_seed(string, bits=seed_bits))
+
+
+def str_to_random_generator(
+        string: str,
+        seed_bits: int = 128
+) -> np.random.Generator:
```
boeddeker, line 58: Do you know, when `np.random.Generator` was introduced?
How about changiing this to `'np.random.Generator'` (i.e. a `str`)?
It would then work for old configurations of python (i.e. less than 3.7, see PEP 563 -- Postponed Evaluation of Annotations) and numpy.

thequilo: Done

boeddeker:
Should we highlight more, that the seed/RandomState/Generator depends on the number of bits?
`This functions outputs a consistent ... dependent on an input string.`
->
```This functions outputs a consistent ... dependent on an input string and the number of bits (`seed_bits`).```

---

Can you add, that `RandomState` only supports up to 32 Bit and that [seeding-and-entropy](https://numpy.org/doc/stable/reference/random/bit_generators/index.html#seeding-and-entropy) recommends to use at least 32 Bit (e.g. don't change that value)?

---

Beside these, this PR LGTM.

boeddeker (approved):

PR #38, opened 2021-01-28
Fix ArrayInterval.sum if emtpy or inverse_mode=True

thequilo:
Previously, `ArrayInterval.sum` was broken for emtpy intervals (`len(interval.intervals) == 0`) and when `interval.inverse_mode = True`. I'm not a 100% sure if `self.shape[0]` is correct.


boeddeker:
Yes, `self.shape` is a tuple and has one entry, so `self.shape[0]` is correct.

boeddeker (approved):
Can you add a doctest for `inverse_mode` (e.g. `ones(10).sum()`)?

PR #40, opened 2021-02-10
Allow export of ArrayIntervals as tuple

TCord:
Enables easily saving ArrayIntervals to and loading them from .json files

### DIFF
--- a/paderbox/array/interval/core.py
+++ b/paderbox/array/interval/core.py
@@ -371,6 +371,7 @@ class ArrayInterval:
         >>> ai.as_tuple
         ('1:4, 5:20, 21:25', (50,))
         """
+        assert self.inverse_mode is False, 'Export of intervals as tuple is only valid for normal mode, not inverse mode!' 
         return self._intervals_as_str, self.shape
 
     @staticmethod
--- a/paderbox/array/interval/core.py
+++ b/paderbox/array/interval/core.py
@@ -362,8 +362,9 @@ class ArrayInterval:
     def as_serializable(self):
         """
         Exports intervals and length of `ArrayInterval` to a serializable object.
-        Intervals are exported as self._intervals_to_str to be human readable. 
-        Allows easy export of ArrayIntervals, e.g. into .json format. 
+        Intervals are exported as `self._intervals_to_str` to be human readable. 
+        Allows easy export of `ArrayIntervals`, e.g. into .json format. 
+        
         >>> from IPython.lib.pretty import pprint
         >>> from paderbox.array.interval.core import ArrayInterval
         >>> ai = ArrayInterval.from_str('1:4, 5:20, 21:25', shape=50)
@@ -383,6 +384,7 @@ class ArrayInterval:
             array: Object of length 2 with items (str: intervals, shape)
         Returns:
             `ArrayInterval` with specified intervals and shape
+            
         >>> from IPython.lib.pretty import pprint
         >>> from paderbox.array.interval.core import ArrayInterval
         >>> at = ('1:4, 5:20, 21:25', 50)
--- a/paderbox/array/interval/core.py
+++ b/paderbox/array/interval/core.py
@@ -358,7 +358,6 @@ class ArrayInterval:
         i_str = ', '.join(i_str)
         return i_str
 
-    @property
     def to_serializable(self):
         """
         Exports intervals and length of `ArrayInterval` to a serializable object.
--- a/paderbox/array/interval/core.py
+++ b/paderbox/array/interval/core.py
@@ -359,7 +359,7 @@ class ArrayInterval:
         return i_str
 
     @property
-    def as_serializable(self):
+    def to_serializable(self):
         """
         Exports intervals and length of `ArrayInterval` to a serializable object.
         Intervals are exported as `self._intervals_to_str` to be human readable. 
@@ -370,7 +370,7 @@ class ArrayInterval:
         >>> ai = ArrayInterval.from_str('1:4, 5:20, 21:25', shape=50)
         >>> ai
         ArrayInterval("1:4, 5:20, 21:25", shape=(50,))
-        >>> ai.as_serializable
+        >>> ai.to_serializable
         ('1:4, 5:20, 21:25', (50,))
         """
         assert self.inverse_mode is False, 'Export of intervals as tuple is only valid for normal mode, not inverse mode!' 
@@ -379,12 +379,14 @@ class ArrayInterval:
     @staticmethod
     def from_serializable(obj):
         """
-        Reverts `as_serializable`.
+        Reverts `to_serializable`.
         Args:
-            array: Object of length 2 with items (str: intervals, shape)
+            obj: Object of length 2 with items (str: intervals, shape)
+            
         Returns:
             `ArrayInterval` with specified intervals and shape
             
+        Example:
         >>> from IPython.lib.pretty import pprint
         >>> from paderbox.array.interval.core import ArrayInterval
         >>> at = ('1:4, 5:20, 21:25', 50)
--- a/paderbox/array/interval/core.py
+++ b/paderbox/array/interval/core.py
@@ -392,7 +392,7 @@ class ArrayInterval:
         ArrayInterval("1:4, 5:20, 21:25", shape=(50,))
         """
         assert len(obj) == 2, f'Expects object of length 2 with items (intervals, shape), got: {obj}' 
-        return ArrayInterval_from_str(array[0], shape=array[1])
+        return ArrayInterval_from_str(obj[0], shape=obj[1])
 
     def __repr__(self):
         if self.inverse_mode:
--- a/paderbox/array/interval/core.py
+++ b/paderbox/array/interval/core.py
@@ -359,33 +359,39 @@ class ArrayInterval:
         return i_str
 
     @property
-    def as_tuple(self):
+    def as_serializable(self):
         """
-        Exports intervals and length of ArrayInterval to a tuple. 
+        Exports intervals and length of ArrayInterval to a serializable object.
+        Reverting 
         Allows easy export of ArrayIntervals, e.g. into .json format. 
         >>> from IPython.lib.pretty import pprint
         >>> from paderbox.array.interval.core import ArrayInterval
         >>> ai = ArrayInterval.from_str('1:4, 5:20, 21:25', shape=50)
         >>> ai
         ArrayInterval("1:4, 5:20, 21:25", shape=(50,))
-        >>> ai.as_tuple
+        >>> ai.as_serializable
         ('1:4, 5:20, 21:25', (50,))
         """
         assert self.inverse_mode is False, 'Export of intervals as tuple is only valid for normal mode, not inverse mode!' 
         return self._intervals_as_str, self.shape
 
     @staticmethod
-    def from_tuple(array):
+    def from_serialized(array):
         """
+        Reverts serialization from 'as_serializable.
+        Args:
+            array: Object of length 2 with items (str: intervals, shape)
+        Returns:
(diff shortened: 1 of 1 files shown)


paderbox/array/interval/core.py
```diff
@@ -358,6 +358,32 @@ def _intervals_as_str(self):
         i_str = ', '.join(i_str)
         return i_str
 
+    @property
+    def as_tuple(self):
+        """
```
thequilo, line 363: Can you add a short description of what this function is useful for?

paderbox/array/interval/core.py
```diff
@@ -358,6 +358,32 @@ def _intervals_as_str(self):
...
+        >>> from IPython.lib.pretty import pprint
+        >>> from paderbox.array.interval.core import ArrayInterval
+        >>> at = ('1:4, 5:20, 21:25', 50)
+        >>> at
+        ('1:4, 5:20, 21:25', 50)
+        >>> ArrayInterval.from_tuple(at)
+        ArrayInterval("1:4, 5:20, 21:25", shape=(50,))
+        """
+        return ArrayInterval_from_str(array[0], shape=array[1])
```
jensheit, line 385: perhaps you should have an assert here, that the length of array is 2

paderbox/array/interval/core.py
TCord: Sure, I thought the name was verbose enough to convey what it's meant for, but I think a description won't hurt.

paderbox/array/interval/core.py
TCord: Done

jensheit (approved):

boeddeker:
I don't like the name `as_tuple`.
I would prefer more something like `to_serializable` (return somthing that can be handled from json/yaml/pickle) or `serialize` (return a str).
The returned value of `as_tuple` is a mixture of both.

Another drawback is, that the user may expect that `as_tuple` returns the interval boundaries as tuple or does the same as `tuple(np.array(array_interval).to_list())`

paderbox/array/interval/core.py
```diff
@@ -358,6 +358,35 @@ def _intervals_as_str(self):
...
+        Allows easy export of ArrayIntervals, e.g. into .json format. 
+        >>> from IPython.lib.pretty import pprint
+        >>> from paderbox.array.interval.core import ArrayInterval
+        >>> ai = ArrayInterval.from_str('1:4, 5:20, 21:25', shape=50)
+        >>> ai
+        ArrayInterval("1:4, 5:20, 21:25", shape=(50,))
+        >>> ai.as_tuple
+        ('1:4, 5:20, 21:25', (50,))
+        """
+        return self._intervals_as_str, self.shape
```
boeddeker, line 374: Can you add an `assert` that `self.inverse_mode` is `False`?

paderbox/array/interval/core.py
```diff
@@ -359,33 +359,39 @@ def _intervals_as_str(self):
...
-        >>> ai.as_tuple
+        >>> ai.as_serializable
         ('1:4, 5:20, 21:25', (50,))
         """
         assert self.inverse_mode is False, 'Export of intervals as tuple is only valid for normal mode, not inverse mode!' 
         return self._intervals_as_str, self.shape
 
     @staticmethod
-    def from_tuple(array):
+    def from_serialized(array):
```
thequilo, line 379: The names `from_serialized` and `to_serializable` don't match. The serialized variant would be a string/bit stream and not a tuple, right?

TCord: Do you think `from_serializable` would be a better name?

thequilo: It would at least match `to_serializable`. Without knowledge about what this function is used for, this name can be confusing. But, I don't have a better idea.

boeddeker: I agree, `from_serializable` would be the counterpart to `to_serializable`.

 - serializable: Something that can be serialized
 - serialize(d): A stream that can be saved, e.g. `str`, `bytes`

paderbox/array/interval/core.py
```diff
@@ -358,6 +358,42 @@ def _intervals_as_str(self):
...
+        >>> ai
+        ArrayInterval("1:4, 5:20, 21:25", shape=(50,))
+        >>> ai.as_serializable
+        ('1:4, 5:20, 21:25', (50,))
+        """
+        assert self.inverse_mode is False, 'Export of intervals as tuple is only valid for normal mode, not inverse mode!' 
+        return self._intervals_as_str, self.shape
+
+    @staticmethod
+    def from_serializable(array):
```
boeddeker, line 379: I think the name `array` is confusing here. I would either call it `serializable` or give it a non-informative name like `obj`.

paderbox/array/interval/core.py
```diff
@@ -358,6 +358,42 @@ def _intervals_as_str(self):
         i_str = ', '.join(i_str)
         return i_str
 
+    @property
+    def as_serializable(self):
+        """
+        Exports intervals and length of ArrayInterval to a serializable object.
+        Reverting
```
boeddeker, line 365: Did you forget to finish this sentence?

paderbox/array/interval/core.py
```diff
@@ -358,6 +358,42 @@ def _intervals_as_str(self):
...
+        >>> from IPython.lib.pretty import pprint
+        >>> from paderbox.array.interval.core import ArrayInterval
+        >>> ai = ArrayInterval.from_str('1:4, 5:20, 21:25', shape=50)
+        >>> ai
+        ArrayInterval("1:4, 5:20, 21:25", shape=(50,))
+        >>> ai.as_serializable
+        ('1:4, 5:20, 21:25', (50,))
+        """
+        assert self.inverse_mode is False, 'Export of intervals as tuple is only valid for normal mode, not inverse mode!' 
+        return self._intervals_as_str, self.shape
```
boeddeker, line 376: Can you add a comment, that we use `self._intervals_as_str` instead of `intervals` to increase the human readability and reduce the size of the serialized object inside a human-readable format like json?

paderbox/array/interval/core.py
thequilo: Also: Backticks for `ArrayInterval`

paderbox/array/interval/core.py
thequilo: Is there an easy way to convert an array with `inverse_mode=True` into one with `inverse_mode=False`? If so, it would be good to have a comment about this in the docstring or error message.

paderbox/array/interval/core.py
```diff
@@ -361,8 +361,8 @@ def _intervals_as_str(self):
     @property
     def as_serializable(self):
         """
-        Exports intervals and length of ArrayInterval to a serializable object.
-        Reverting 
+        Exports intervals and length of `ArrayInterval` to a serializable object.
+        Intervals are exported as self._intervals_to_str to be human readable. 
         Allows easy export of ArrayIntervals, e.g. into .json format.
```
thequilo, line 366: Backticks for `ArrayInterval`s

thequilo: And `self._intervals_to_str`

paderbox/array/interval/core.py
```diff
@@ -358,6 +358,42 @@ def _intervals_as_str(self):
...
+        Allows easy export of ArrayIntervals, e.g. into .json format. 
+        >>> from IPython.lib.pretty import pprint
+        >>> from paderbox.array.interval.core import ArrayInterval
+        >>> ai = ArrayInterval.from_str('1:4, 5:20, 21:25', shape=50)
+        >>> ai
+        ArrayInterval("1:4, 5:20, 21:25", shape=(50,))
+        >>> ai.as_serializable
+        ('1:4, 5:20, 21:25', (50,))
+        """
+        assert self.inverse_mode is False, 'Export of intervals as tuple is only valid for normal mode, not inverse mode!'
```
thequilo, line 375: Line too long?

paderbox/array/interval/core.py
```diff
@@ -358,6 +358,42 @@ def _intervals_as_str(self):
...
+        assert self.inverse_mode is False, 'Export of intervals as tuple is only valid for normal mode, not inverse mode!' 
+        return self._intervals_as_str, self.shape
+
+    @staticmethod
+    def from_serializable(obj):
+        """
+        Reverts `as_serializable`.
+        Args:
+            array: Object of length 2 with items (str: intervals, shape)
+        Returns:
```
thequilo, line 384: The Google docstring format recommends newlines between the sections like "Args" and "Returns". It also recommends an "Examples" section for the examples/doctests, but not everyone likes this.

thequilo: And rename "array" -> "obj" in docstring

paderbox/array/interval/core.py
thequilo: And a newline between description and examples/doctests helps the readability

paderbox/array/interval/core.py
```diff
@@ -358,6 +358,42 @@ def _intervals_as_str(self):
         i_str = ', '.join(i_str)
         return i_str
 
+    @property
+    def as_serializable(self):
```
thequilo, line 362: You could add a type annotation, at least `tuple`, or `Tuple[str, Tuple[int]]`

paderbox/array/interval/core.py
```diff
@@ -358,6 +358,42 @@ def _intervals_as_str(self):
...
+        >>> ai
+        ArrayInterval("1:4, 5:20, 21:25", shape=(50,))
+        >>> ai.as_serializable
+        ('1:4, 5:20, 21:25', (50,))
+        """
+        assert self.inverse_mode is False, 'Export of intervals as tuple is only valid for normal mode, not inverse mode!' 
+        return self._intervals_as_str, self.shape
+
+    @staticmethod
+    def from_serializable(obj):
```
thequilo, line 380: Type annotations could help to understand this method.

paderbox/array/interval/core.py
```diff
@@ -358,6 +358,42 @@ def _intervals_as_str(self):
         i_str = ', '.join(i_str)
         return i_str
 
+    @property
```
thequilo, line 361: Why is this a property? `as_something` is often a method (e.g., `dataclasses.as_dict`) and could be confusing if this one is a property.

paderbox/array/interval/core.py
boeddeker: It depends, when the shape is known, you can convert it. Otherwise, it is not possible, i.e. the last interval goes to infinity.

paderbox/array/interval/core.py
boeddeker: For asserts is a single line better, it improves the Exception.

paderbox/array/interval/core.py
thequilo: Or call this `to_serializable` as in `np.ndarray.to_list`?

paderbox/array/interval/core.py
TCord: It does not only work with a `tuple`, so I am unsure if it's good to add a type annotation

paderbox/array/interval/core.py
TCord: I wanted to keep it as similar as possible to the property `_intervals_to_str`. I renamed it to `to_serializable`, now.

boeddeker: Do you also remove the `@property`? The `@property` for `_intervals_to_str` is only ok, because it is private.

boeddeker (approved):

paderbox/array/interval/core.py
TCord: Done

PR #41, opened 2021-02-11
Allow assigning ArrayInterval to ArrayInterval and improvements to Exceptions of ArrayInterval

thequilo:


paderbox/array/interval/core.py
```diff
@@ -466,33 +474,66 @@ def __setitem__(self, item, value):
...
-        elif isinstance(value, (tuple, list, np.ndarray)):
-            assert len(value) == stop - start, (start, stop, stop - start, len(value), value)
-            ai = ArrayInterval(value, inverse_mode=self.inverse_mode)
-            intervals = self.intervals
-            intervals = cy_non_intersection((start, stop), intervals)
-            self.intervals = intervals + tuple([(s + start, e + start) for s, e in ai.intervals])
+                self.intervals = cy_non_intersection((start, stop), self.intervals)
+        elif isinstance(value, (tuple, list, np.ndarray, ArrayInterval)):
+            if not isinstance(value, ArrayInterval):
+                value = ArrayInterval(value, inverse_mode=self.inverse_mode)
```
boeddeker, line 517: Why `inverse_mode=self.inverse_mode` and not `inverse_mode=False`?

boeddeker, line 517: Can you add a comment, why `inverse_mode=self.inverse_mode`?

paderbox/array/interval/core.py
```diff
@@ -466,33 +474,66 @@ def __setitem__(self, item, value):
...
+
+            if len(value) != stop - start:
+                raise ValueError(
+                    f'Could not broadcast input with length {len(value)} into '
+                    f'shape {stop - start}'
+                )
+            if value.inverse_mode != self.inverse_mode:
+                raise ValueError(
```
boeddeker, line 525: Maybe a `NotImplementedError`?

paderbox/array/interval/core.py
```diff
@@ -514,16 +555,33 @@ def __getitem__(self, item):
...
+        >>> ai[-25:-20]
+        array([ True,  True,  True,  True,  True])
+        >>> ai[-1:]
+        array([])
+
         """
         if isinstance(item, (int, np.integer)):
+            index = item
+            if index < 0:
+                index = index + self.shape[0]
```
boeddeker, line 567: `.shape[0]` -> `.shape[-1]` Just for the case, we support independent dims in the furture

boeddeker: Shape may be None. Can you catch this case?

paderbox/array/interval/core.py
```diff
@@ -466,33 +474,83 @@ def __setitem__(self, item, value):
...
+                f'{self.__class__.__name__}.__setitem__ only supports slices '
+                f'for indexing, not {item}'
+            )
 
         start, stop = cy_parse_item(item, self.shape)
 
         if np.isscalar(value):
+            if value not in (0, 1):
+                # Numpy interprets values as boolean even if they are larger
+                # than 1, why doesn't the ArrayInterval do the same?
```
boeddeker, line 519: Checking if the value is a Boolean (or 0, 1) is more verbose, and I don't know a case, where other values are useful. Usually they indicate a bug or a wrong assumption of the user.

paderbox/array/interval/core.py
thequilo: Done

paderbox/array/interval/core.py
thequilo: Done

paderbox/array/interval/core.py
thequilo: Done

boeddeker (approved):

paderbox/array/interval/core.py
thequilo: I updated the comment

PR #42, opened 2021-04-30
Add partial_decorator

thequilo:
`partial_decorator` allows calling a function multiple times to get a behavior similar to `functools.partial` (See doctests for examples).

Points that might need discussion:
 - Naming of the decorator
 - Naming of the file (`pb.utils.functional`) and potential imports on higher levels
 - Naming of the `nested` argument and its default value
 - `TypeErrors` for unexpected arguments don't contain the function name (because `inspect.Signature.bind` doesn't add it to the exception message)


thequilo:
The failing test cases are not caused by my changes. Something seems to have changed in the `test_new_sub_dir` random stuff.

boeddeker (approved):

jensheit (approved):

PR #43, opened 2021-05-04
Add miscellaneous utility functions

thequilo:
This PR adds some utility functions from my personal toolbox that I think could be useful for others.

Points for discussion:
 - Naming of `get_by_path` and `set_by_path` could also be `nested_get` and `nested_set`, but I kept the "development names" for now. `nested_set` could be confused for a datatype, a `set` that contains `set`s.

### DIFF
--- a/paderbox/utils/misc.py
+++ b/paderbox/utils/misc.py
@@ -5,30 +5,9 @@ find something useful if you alter it a little bit.
 
 """
 
-import collections
 import os
 import sys
-
-
-def update_dict(d, u):
-    """ Recursively update dict d with values from dict u.
-
-    Args:
-        d: Dict to be updated
-        u: Dict with values to use for update
-
-    Returns: Updated dict
-
-    """
-    for k, v in u.items():
-        if isinstance(v, collections.Mapping):
-            default = v.copy()
-            default.clear()
-            r = update_dict(d._get(k, default), v)
-            d[k] = r
-        else:
-            d[k] = v
-    return d
+from typing import Iterable, Hashable
 
 
 def interleave(*lists):
@@ -74,3 +53,67 @@ class PrintSuppressor:
     def __exit__(self, exc_type, exc_val, exc_tb):
         sys.stdout.close()
         sys.stdout = self._original_stdout
+
+
+def all_equal(x: Iterable[Hashable]) -> bool:
+    """
+    Checks if all elements in `x` are equal. Returns `False` if `x` is empty.
+
+    Defined to improve readability.
+
+    Examples:
+        >>> all_equal((1, 1, 1, 1))
+        True
+        >>> all_equal((1, 1, 1, 2))
+        False
+    """
+    return len(set(x)) == 1
+
+
+def all_unique(x: Iterable[Hashable]) -> bool:
+    """
+    Checks if all elements in `x` are unique. Returns `True` if `x` is empty.
+
+    Defined to improve readability.
+
+    Examples:
+        >>> all_unique((1, 2, 3, 4))
+        True
+        >>> all_unique((1, 2, 3, 1))
+        False
+    """
+    return len(set(x)) == len(list(x))
+
+
+def all_in(x: Iterable[Hashable], y: Iterable[Hashable]) -> bool:
+    """
+    Check if all elements in `x` are in `y`. Returns `True` if `x` is empty.
+
+    Equivalent to `set(x).issubset(y)`, but faster.
+
+    Defined to improve readability.
+
+    Examples:
+        >>> all_in([1, 2, 2, 1, 2], [1, 2, 3])
+        True
+        >>> all_in([1, 2, 2, 4, 2], [1, 2, 3])
+        False
+    """
+    return all(x_ in y for x_ in x)
+
+
+def any_in(x: Iterable[Hashable], y: Iterable[Hashable]) -> bool:
+    """
+    Check if any elements in `x` is in `y`. Returns `True` if `x` is empty.
+
+    Defined to improve readability.
+
+    Examples:
+        >>> any_in([1, 2, 2, 1, 2], [1, 2, 3])
+        True
+        >>> any_in([1, 2, 2, 4, 2], [1, 2, 3])
+        True
+        >>> any_in([1, 2, 2, 4, 2], [3, 5, 6])
+        False
+    """
+    return any(x_ in y for x_ in x)
--- a/paderbox/utils/nested.py
+++ b/paderbox/utils/nested.py
@@ -2,14 +2,17 @@ import copy
 
 import collections
 import itertools
+from typing import Optional, Any, Union, Sequence, Mapping, Tuple, Callable
 
 
-def flatten(d, sep='.', flat_type=dict):
-    """Flatten a nested dict using a specific separator.
+def flatten(d, sep: Optional[str] = '.', *, flat_type=dict):
+    """
+    Flatten a nested `dict` using a specific separator.
 
-    :param sep: When None, return dict with tuple keys (guaranties inversion of
-                flatten) else join the keys with sep
-    :param flat_type: Allow other mappings instead of flat_type to be
+    Args:
+        sep: When `None`, return `dict` with `tuple` keys (guarantees inversion
+                of flatten) else join the keys with sep
+        flat_type:  Allow other mappings instead of `flat_type` to be
                 flattened, e.g. using an isinstance check.
 
     import collections
@@ -31,6 +34,7 @@ def flatten(d, sep='.', flat_type=dict):
     c_b_y 10
     d [1, 2, 3]
     """
+
     # https://stackoverflow.com/a/6027615/5766934
 
     # {k: v for k, v in d.items()}
@@ -54,8 +58,15 @@ def flatten(d, sep='.', flat_type=dict):
         }
 
 
-def deflatten(d, sep='.', maxdepth=-1):
-    """Build a nested dict from a flat dict respecting a separator.
+def deflatten(d: dict, sep: Optional[str] = '.', maxdepth: int = -1):
+    """
+    Build a nested `dict` from a flat dict respecting a separator.
+
+    Args:
+        d: Flattened `dict` to reconstruct a `nested` dict from
+        sep: The separator used in the keys of `d`. If `None`, `d.keys()` should
+            only contain `tuple`s.
+        maxdepth: Maximum depth until wich nested conversion is performed
 
     >>> d_in = {'a': 1, 'c': {'a': 2, 'b': {'x': 5, 'y' : 10}}, 'd': [1, 2, 3]}
     >>> d = flatten(d_in)
@@ -137,10 +148,12 @@ def nested_update(orig, update):
 def nested_merge(default_dict, *update_dicts, allow_update=True, inplace=False):
     """
     Nested updates the first dict with all other dicts.
-    The last dict has the highest priority when allow_update is True.
-    When allow_update is False, it is assumed, that no values gets overwritten.
 
-    When inplace is True, the default_dict is manipulated inplace. This is
+    The last dict has the highest priority when `allow_update` is `True`.
+    When `allow_update` is `False`, it is assumed that no values get
+    overwritten and an exception is raised when duplicate keys are found.
+
+    When `inplace` is `True`, the `default_dict` is manipulated inplace. This is
     useful for `collections.defaultdict`.
 
     # Example from: https://stackoverflow.com/q/3232943/5766934
@@ -239,13 +252,13 @@ def nested_op(
         keep_type=True,
         mapping_type=collections.abc.Mapping,
         sequence_type=(tuple, list),
-
 ):
-    """This function is `nested_map` with a fancy name.
-
-    Applies the function "func" to the leafs of the nested data structures.
+    """
+    Applies the function `func` to the leafs of the nested data structures in
+    `arg1` and `*args`.
     This is similar to the map function that applies the function the the
     elements of an iterable input (e.g. list).
+    This function is `nested_map` with a fancy name.
 
     CB: Should handle_dataclass be True or False?
         Other suggestions for the name "handle_dataclass"?
@@ -370,7 +383,7 @@ def nested_op(
                       if hasattr(arg, '__dataclass_fields__')
                       else arg
                       for arg in args
-                    ],
+                      ],
                     broadcast=broadcast,
                     mapping_type=mapping_type,
                     sequence_type=sequence_type,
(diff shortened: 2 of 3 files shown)


paderbox/utils/misc.py
```diff
@@ -74,3 +53,67 @@ def __enter__(self):
...
+
+    Defined to improve readability.
+
+    Examples:
+        >>> all_equal((1, 1, 1, 1))
+        True
+        >>> all_equal((1, 1, 1, 2))
+        False
+    """
+    return len(set(x)) == 1
```
boeddeker, line 70: How about checking, that the input is not a dict/mapping?

paderbox/utils/misc.py
```diff
@@ -74,3 +53,67 @@ def __enter__(self):
...
+
+    Defined to improve readability.
+
+    Examples:
+        >>> all_unique((1, 2, 3, 4))
+        True
+        >>> all_unique((1, 2, 3, 1))
+        False
+    """
+    return len(set(x)) == len(list(x))
```
boeddeker, line 85: How about checking, that the input is not a dict/mapping?

paderbox/utils/misc.py
```diff
@@ -74,3 +53,67 @@ def __enter__(self):
...
+
+
+def all_in(x: Iterable[Hashable], y: Iterable[Hashable]) -> bool:
```
boeddeker, line 88: Is the name `issubset` more obvious?

paderbox/utils/misc.py
```diff
@@ -74,3 +53,67 @@ def __enter__(self):
...
+    """
+    Check if all elements in `x` are in `y`. Returns `True` if `x` is empty.
+
+    Equivalent to `set(x).issubset(y)`, but faster.
```
boeddeker, line 92: This depends. I expected that `set(x).issubset(y)` is in relevant cases faster (i.e. large x and large y and return is True).

paderbox/utils/misc.py
```diff
@@ -74,3 +53,67 @@ def __enter__(self):
...
+
+    Defined to improve readability.
+
+    Examples:
+        >>> all_in([1, 2, 2, 1, 2], [1, 2, 3])
+        True
+        >>> all_in([1, 2, 2, 4, 2], [1, 2, 3])
+        False
+    """
+    return all(x_ in y for x_ in x)
```
boeddeker, line 102: Is your assumption that this function should usually return False? Otherwise, a list comprehension is faster

paderbox/utils/misc.py
```diff
@@ -74,3 +53,67 @@ def __enter__(self):
...
+
+
+def any_in(x: Iterable[Hashable], y: Iterable[Hashable]) -> bool:
```
boeddeker, line 105: Is this not an intersection check?

paderbox/utils/nested.py
```diff
@@ -400,3 +413,196 @@ def squeeze_nested(orig):
...
+        *,
+        delimiter: str = '.',
+) -> None:
+    """
+    Sets a value in the nested dictionary `d` by the dotted path.
+
+    Modifies `d` inplace.
+
+    Args:
+        d: The container to get the value from
```
boeddeker, line 510: Forgot to adjust the text?

paderbox/utils/nested.py
```diff
@@ -400,3 +413,196 @@ def squeeze_nested(orig):
...
+            ) from e
+    return d
+
+
+def get_by_path(
+        d: Union[Mapping, Sequence],
+        path: Union[str, Tuple[Any, ...], None],
+        *,
+        broadcast: bool = False,
+        delimiter: str = '.',
```
boeddeker, line 449: In flatten and deflatten we use `sep`.
The python doc contains the signature `str.split(sep=None, maxsplit=-1)` (https://docs.python.org/3/library/stdtypes.html#str.split).

How about to use also `sep` here?

paderbox/utils/nested.py
```diff
@@ -400,3 +413,196 @@ def squeeze_nested(orig):
...
+        2
+        >>> get_by_path(d, 'a.b.c', broadcast=True)
+        'b'
+        >>> get_by_path(d, 'c.b.c', default=42)
+        42
+    """
+    if path is None:
+        return d
+    if isinstance(path, str):
+        path = path.split(delimiter)
```
boeddeker, line 488: Move this code to the helper? This code is in both function and `_get_by_path` is not recursive.

paderbox/utils/nested.py
```diff
@@ -400,3 +413,196 @@ def squeeze_nested(orig):
...
+
+def _get_by_path(
+        d: Union[Mapping, Sequence],
+        path: Tuple[Any, ...],
+        broadcast: bool = False
+):
+    """ Helper function for `get_by_path` and `set_by_path`. """
+    for k in path:
+        try:
+            d = d[k]
```
boeddeker, line 426: For broadcast you have to catch much more cases. The exception can have any type.

paderbox/utils/nested.py
```diff
@@ -400,3 +413,196 @@ def squeeze_nested(orig):
...
+                d, k
+            ) from e
+    return d
+
+
+def get_by_path(
+        d: Union[Mapping, Sequence],
+        path: Union[str, Tuple[Any, ...], None],
+        *,
+        broadcast: bool = False,
```
boeddeker, line 448: I don't like the name `broadcast`. But I have no better name, so it is probably fine.

paderbox/utils/misc.py
thequilo: I checked again and, at least for integers and strings, `set(s).issubset(y)` is always faster. I don't know what I did when I wrote the function and the comment. I'll change it to `set(s).issubset(y)`.

paderbox/utils/misc.py
thequilo: Could be, in some cases, and in others, it's not. This really depends on the use case, but you might be correct for most cases.

paderbox/utils/misc.py
thequilo: Yes, it is, do you think `intersects_with` is a better name? This could be very confusing

jensheit: I like the idea of keeping the names of these functions similar. I would keep the any/all prefix

paderbox/utils/nested.py
```diff
@@ -400,3 +413,196 @@ def squeeze_nested(orig):
...
+    Examples:
+        >>> d = {'a': 'b', 'c': {'d': {'e': 'f'}, 'g': [1, [2, 3], 4]}}
+        >>> get_by_path(d, 'a')
+        'b'
+        >>> get_by_path(d, 'c.d.e')
+        'f'
+        >>> get_by_path(d, ('c', 'g', 1, 0))
+        2
+        >>> get_by_path(d, 'a.b.c', broadcast=True)
+        'b'
```
jensheit, line 481: okay, that is not what I expected from broadcast. Isnt that something like early stopping?

paderbox/utils/nested.py
```diff
@@ -400,3 +413,196 @@ def squeeze_nested(orig):
...
+        >>> get_by_path(d, 'c.b.c', default=42)
+        42
+    """
+    if path is None:
+        return d
+    if isinstance(path, str):
+        path = path.split(delimiter)
+    try:
+        return _get_by_path(d, path, broadcast=broadcast)
```
jensheit, line 490: do you realy need a subfunction here? Are you using it somewhere else?

paderbox/utils/nested.py
thequilo: Yes, it basically is early stopping. But isn't broadcasting also early stopping? The value of "a.b" is here "broadcasted" to any path in the sub-container indexed by "a.b".

paderbox/utils/misc.py
boeddeker: You use a generator comprehension in combination with `all`.
When one of the first entries is False, this is usually faster (i.e. expected output is `False`).
Maybe you had such a case in the past.

paderbox/utils/misc.py
boeddeker: My idea was to rename both function :).
I am not sure, that `intersects_with` is the best naming. It sounds like a non-symmetric operation.
Maybe something like `has_intersection`?

thequilo: I chose the names `all_in` and `any_in` as extensions of Python's `in` operator, so that, for example,
```python
assert [o in available_options for o in options]
```
becomes more obvious
```python
assert all_in(options, available_options)
```
I don't think that `has_intersection` is more obvious in this case. But there are other cases where it is.

boeddeker: I ask just because, I prefer to use existing python names, when they exist.
Since `all_in` is `set.issubset`, where we allow more types, I thought it would be better to align the names.

For `any_in` it is a little different, because it needs two functions (`set.intersection` and `bool`).

I think in your example, both are easy to interpret:

`assert all_in(options, available_options)`

vs

`assert issubset(options, available_options)`

paderbox/utils/nested.py
```diff
@@ -400,3 +413,295 @@ def squeeze_nested(orig):
...
     return orig
+
+
+def get_by_path(
+        container: Union[Mapping, Sequence],
+        path: Union[str, Tuple[Any, ...], None],
+        *,
+        allow_early_stopping: bool = False,
+        sep: str = '.',
+        default: Any = ...,
```
boeddeker, line 424: Why do you use `Ellipsis` here?
I think you wanted to use something like:

```
NO_VALUE = object()
def get_by_path(..., default=NO_VALUE): ...
```

paderbox/utils/nested.py
```diff
@@ -400,3 +413,295 @@ def squeeze_nested(orig):
...
+        >>> container
+        {'a': {'b': {'c': [1, 2, 42], 'd': 'e'}}}
+    """
+    if isinstance(path, str):
+        path = path.split(sep)
+    container = get_by_path(container, path[:-1])
+    container[path[-1]] = value
+
+
+def nested_iter_items(container, *, iter_types=(dict, tuple, list)):
```
boeddeker, line 531: Maybe adding a comment, that this function could get a `depth` or `max_depth` argument in the future, when someone needs it??

paderbox/utils/nested.py
```diff
@@ -400,3 +413,295 @@ def squeeze_nested(orig):
...
+
+    def get(self, item, default=..., *, sep=None):
+        if sep is None:
+            sep = self.sep
+        return get_by_path(
+            self._container, item, allow_early_stopping=self.allow_partial_path,
+            default=default, sep=sep,
+        )
+
+    def items(self):
```
boeddeker, line 621: You could define keys (and values) as:

```python
def keys(self):
    return (k for k, _ in self.items())
```
(or using operator itemgetter)

paderbox/utils/misc.py
thequilo: Done

paderbox/utils/misc.py
thequilo: Done

paderbox/utils/misc.py
thequilo: Changed it to set operations

paderbox/utils/nested.py
thequilo: Done

paderbox/utils/nested.py
thequilo: Moved the helper into `get_by_path`

paderbox/utils/nested.py
thequilo: No, I removed it

paderbox/utils/nested.py
thequilo: Yes, I saw the `Ellipsis` being used for this somewhere (I don't remember where), but your suggestion is better in case anyone want's to have `Ellipsis` as default. Should we define a global `paderbox.NO_VALUE` constant?

paderbox/utils/nested.py
thequilo: Usually, the `keys` function of a mapping returns a list or similar instead of a generator. Not sure if this can be confusing. We can also define `__len__` by `len(list(self.items()))`, but I don't want to run through the whole nested container for this

boeddeker: Partial. I think in python2 keys, values and items yielded lists, but in python 3 they started to return special types, and now you mostly cast them to a list or tuple to work with them.

My motivation is more: When  `items` is defined, I expect that `keys` and `values` exists.
I agree, that all of them aren't so well-defined for this type.

paderbox/utils/nested.py
boeddeker: Ups, I forgot the `_` in `_NO_VALUE`.

I think it is common to define the `NO_VALUE` as a hidden variable, in this way it is ensured, that no one can use it wrong. I checked `inspect` and the have only `inspect.Parameter.empty` as public (i.e. `inspect._empty` and `inspect._void` are private)

paderbox/utils/nested.py
thequilo: I agree with you, I also expect `keys` and `values` to exist when `items` exists, but the definition of `keys` and `items` is different from a typical `Mapping`.  But since we can't subclass `Mapping` because we can't define `__len__`, I think it's fine.

boeddeker: Following your arguments, keys is not well-defined, hence we should not provide `items`?
With the definition of items, you fixed everything. The only problem is, that execution time of `__len__` is not constant (`keys`, `values` and `items` is constant, because the computation happens, when you iterate.).

thequilo: Yes, that's right. So you say, we should also provide `__len__`?

Another problem is that `items` yields the leaves and not all possible keys you can index the `FlatView` with.

paderbox/utils/nested.py
thequilo: I just realized that the private `_NO_VALUE` constant makes it hard to wrap this function without accessing the private module member, so it might be better to have it public.

paderbox/utils/nested.py
boeddeker: Here, the things that I think have arguments (I am fine with the first three, with a small preference of the first):
 - Provide None
 - Provide `items`, `values`, `keys`
 - Provide `items`, `values`, `keys`, `__len__`
 - Provide `values` (Difficult, because values should not contain duplicates)

My argumentation is that I cannot find an argument, to only implement `items`.
It is implemented as a generator, and `values` and `keys` can have the same implementation (My example was a generator comprehension) and `operator.itemgetter` is also lazy.

An implementation of `__len__` is optional, because it is relative expensive, while it shouldn't be the case for an implementation of `__len__` (We can compensate it with caching).

My small preference to provide non, is the fact, that I don't see the use case and the non-unique definition (Once we implement it, a change of the definition is a breaking change).

boeddeker (approved):

paderbox/utils/nested.py
thequilo: > An implementation of __len__ is optional, because it is relative expensive, while it shouldn't be the case for an implementation of __len__ (We can compensate it with caching).

Unfortunately we can't use caching: we can't detect changes in the wrapped container, so we don't know when to reset the cache

My argument for `items` is that I implemented `nested_iter_items` and had some use-cases in the past where I needed something like this. I would be confused if `FlatView` wouldn't support the same functionality in some way and `items` is the most obvious place for it.

We could also implement an `items` which iterates over everything and an `leaf_items` that only iterates over the leaves.

PR #47, opened 2021-05-20
Add Github Actions for test cases

mdeegen:


.github/workflows/tests.yml
```diff
@@ -0,0 +1,53 @@
...
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.6, 3.7]
```
boeddeker, line 15: Can you add 3.8?

mdeegen: In the azure-pipelines.yml it said "Python 3.8 does not work at the moment." so i did not add 3.8.
But if you say i should add it here i can do that, sure.

jensheit:
the combination of two independent changes is a bit confusing here, but overall it seems to be working

jensheit (approved):

boeddeker (approved):
Thank you. Looks good to me.

PR #48, opened 2021-05-31
Add activity plot function

thequilo:
This is a rough draft of the function. There are many things that could need discussion or improvement:
 - naming of the functions and their parameters
 - How to pass through kwargs to lower-level functions
 - Is a `dict` of `ArrayIntervals` the correct representation? Would it be easier to also support lists of tuples of start and end times? or a tuple of two lists of start and end times? The raw json contains a list of start times and a list of end times
 - Generalize hbar even more so that other args like color are broadcasted to the intervals in the `ArrayInterval`s?
 - Is this the correct location for the `activity` function or is it too specific? Would it be better located elsewhere, e.g., next to the database?

### DIFF
--- a/paderbox/visualization/plot.py
+++ b/paderbox/visualization/plot.py
@@ -965,8 +965,11 @@ def activity(
             `speaker_labels`. The activity of a single speaker can be an
             `pb.array.interval.ArrayInterval` or a list of tuples of start and
             stop times (`[(start1, stop1), (start2, stop2), ...]`).
-        speaker_labels: Speaker labels. Has to be `None` when
-            `activity_intervals` is a `dict`.
+        speaker_labels: Speaker labels. If it is `None` and `activity_intervals`
+            is a `dict`, `speaker_labels` is set to `activity_intervals.keys()`.
+            If `activity_intervals` is a `dict` and `speaker_labels` is given,
+            `speaker_labels` are used to select speakers from
+            `activity_intervals`.
         segment_boundary_intervals: Intervals that indicate boundaries of
             segments. These are drawn with alpha channel. The format has to be
             the same as `activity_intervals`.
--- a/paderbox/visualization/plot.py
+++ b/paderbox/visualization/plot.py
@@ -23,8 +23,6 @@ __all__ = [
     'image'
 ]
 
-from paderbox.utils.functional import partial_decorator
-
 
 def create_subplot(f):
     """ This decorator creates a subplot and passes the axes object if needed.
@@ -856,7 +854,7 @@ def seq2seq_alignment(alignment, targets=None, decode=None, ax=None,
 
 
 @create_subplot
-def barh(y, width, height=0.8, left=None, *, ax=None, scale=1, **kwargs):
+def barh(y, width, left=None, *, ax=None, **kwargs):
     """
     A wrapper around `plt.barh` that can also handle
     `pb.array.interval.ArrayInterval`s.
@@ -867,12 +865,9 @@ def barh(y, width, height=0.8, left=None, *, ax=None, scale=1, **kwargs):
             element of `width` has its own y coordinate)
         width: Scalar, array-like, `pb.array.interval.ArrayInterval` or list of
             `pb.array.interval.ArrayInterval`.
-        height:
-        left:
-        ax:
-        scale:
+        left: Starting points of the bars. If `width` is an `ArrayInterval`,
+            `left` has to be `None`.
         **kwargs:
-
     """
     import paderbox as pb
     from paderbox.array.interval import ArrayInterval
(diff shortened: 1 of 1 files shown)


paderbox/visualization/plot.py
```diff
@@ -846,3 +853,197 @@ def seq2seq_alignment(alignment, targets=None, decode=None, ax=None,
...
+    Args:
+        y: The `y` coordinate(s). Can be a scalar (all bars are drawn at the
+            same y coordinate) or array-like of length `len(width)` (each
+            element of `width` has its own y coordinate)
+        width: Scalar, array-like, `pb.array.interval.ArrayInterval` or list of
+            `pb.array.interval.ArrayInterval`.
+        height:
+        left:
+        ax:
+        scale:
```
jensheit, line 873: should we copy the args definition from plt.barh?

paderbox/visualization/plot.py
```diff
@@ -846,3 +853,197 @@ def seq2seq_alignment(alignment, targets=None, decode=None, ax=None,
...
+        width = new_width
+        y = new_y
+    elif not isinstance(width, Iterable):
+        width = [width]
+    ax.barh(y, width, height, left, **kwargs)
+    return ax
+
+
+@create_subplot
+def axvspan(start, stop=None, ax=None, **kwargs):
```
jensheit, line 922: this code does not work with the default ax, so either we have to change the default or reset it to plt if ax=None

paderbox/visualization/plot.py
```diff
@@ -846,3 +853,197 @@ def seq2seq_alignment(alignment, targets=None, decode=None, ax=None,
...
+        activity_intervals:
+        segment_boundary_intervals:
+        ax:
+        activity_kwargs:
+        overlap_kwargs:
+        total_activity_kwargs:
+        scale:
+        bar_edgecolor:
+        plot_total_activity:
+
```
jensheit, line 973: I think you need to add some explanation here, at least for activity_intervals and segment_boundary_intervals

jensheit: y_labels is missing

paderbox/visualization/plot.py
```diff
@@ -846,3 +853,197 @@ def seq2seq_alignment(alignment, targets=None, decode=None, ax=None,
...
+    Returns:
+
+    """
+    import numpy as np
+    import paderbox as pb
```
jensheit, line 978: do we really need to import those to inside of a function?

paderbox/visualization/plot.py
```diff
@@ -846,3 +853,197 @@ def seq2seq_alignment(alignment, targets=None, decode=None, ax=None,
...
+    ))
+    num_active = np.zeros((length,), dtype=np.int)
+    for a in activity_intervals:
+        num_active += a[:length]
+
+    # Plot overlaps
+    if plot_overlaps:
+        kwargs = {'color': 'orange', 'alpha': 0.4}
+        if overlap_kwargs is not None:
+            kwargs.update(overlap_kwargs)
```
jensheit, line 1004: sollte man hier wirklich die standard kwargs verändern?

paderbox/visualization/plot.py
```diff
@@ -846,3 +853,197 @@ def seq2seq_alignment(alignment, targets=None, decode=None, ax=None,
...
+@create_subplot
+def activity(
+        activity_intervals,
+        y_labels=None,
+        segment_boundary_intervals=None,
+        ax=None,
+        activity_kwargs=None,
+        overlap_kwargs=None,
+        total_activity_kwargs=None,
+        plot_total_activity=True,
```
jensheit, line 956: I would move this option in front of the kwargs
Am I right to assume, that this plots activity bar if any speaker is active?

paderbox/visualization/plot.py
thequilo: Yes, we could do that. Or write "see plt.barh".

paderbox/visualization/plot.py
thequilo: It works with the default `ax`. It gets filled by the `create_subplot` decorator, as for all other functions in this file

paderbox/visualization/plot.py
thequilo: No, in my private repository, where I copied the function from, this was the only function that used np and pb, but here other functions also use np.

paderbox/visualization/plot.py
thequilo: `plot_total_acitivty` plots an additional bar for the overall activity (green bar in the image)
image

thequilo:
Well, something went wrong with the rebase on master

paderbox/visualization/plot.py
jensheit: ah I didnt see the decorator you are right of course

jensheit (approved):

PR #49, opened 2021-06-08
Fix ArrayInterval.__reduce__ for inverse mode

thequilo:
Previously, `ArrayInterval.__reduce__` ignored `inverse_mode`.

### DIFF
--- a/paderbox/array/interval/core.py
+++ b/paderbox/array/interval/core.py
@@ -30,7 +30,6 @@ def ArrayInterval_from_str(string, shape):
     """
     ai = zeros(shape)
     if string == '':
-        print('empty interval found')
         pass
     else:
         if not ',' in string:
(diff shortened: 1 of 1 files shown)


paderbox/array/interval/core.py
```diff
@@ -274,6 +273,12 @@ def __array__(self, dtype=np.bool):
         else:
             return self[:]
 
+    @staticmethod
+    def _reduce_restore(intervals_as_str, shape, inverse_mode):
+        interval = ArrayInterval.from_str(intervals_as_str, shape)
```
boeddeker, line 278: Should we add the `inverse_mode` flag to `from_str`?

thequilo: Yes, we could do that

boeddeker (approved):

PR #50, opened 2021-06-14
Normalize and validate shape of ArrayInterval while setting it

thequilo:
Before, it was possible to do
```python
ai = zeros('a')
```
without an exception. Now it raises an exception and also checks the shape on assignment like `ai.shape = (1, 2, 3, 4)`.

### DIFF
--- a/paderbox/array/interval/core.py
+++ b/paderbox/array/interval/core.py
@@ -46,9 +46,13 @@ def intervals_to_str(intervals):
     return ', '.join(f'{start}:{end}' for start, end in intervals)
 
 
-def _validate_shape(shape):
-    if shape is None or isinstance(shape, int):
-        return
+def _normalize_shape(shape):
+    if shape is None:
+        return None
+
+    if isinstance(shape, int):
+        return shape,
+
     if not isinstance(shape, (tuple, list)):
         raise TypeError(f'Invalid shape {shape} of type {type(shape)}')
     if not len(shape) == 1:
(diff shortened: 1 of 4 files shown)


paderbox/array/interval/core.py
```diff
@@ -46,6 +46,27 @@ def intervals_to_str(intervals):
     return ', '.join(f'{start}:{end}' for start, end in intervals)
 
 
+def _validate_shape(shape):
+    if shape is None or isinstance(shape, int):
+        return
+    if not isinstance(shape, (tuple, list)):
+        raise TypeError(f'Invalid shape {shape} of type {type(shape)}')
+    if not len(shape) == 1:
+        raise ValueError(f'Invalid shape {shape} has to have length 1')
```
boeddeker, line 55: Can you add a comment to the source code?
Something like:
`As of now, we only support 1D. We haven't decided, how to handle what higher shapes mean. Probably the last will be, what it is now and the earlier dims will be independent dims.`

paderbox/array/interval/core.py
```diff
@@ -46,6 +46,27 @@ def intervals_to_str(intervals):
...
+    if not isinstance(shape[0], int):
```
boeddeker, line 56: For long term: `0` -> `-1`

boeddeker:
Can you combine `_validate_shape` and `_normalize_shape`?
I don't see a reason to split them.

boeddeker (approved):

PR #51, opened 2021-07-12
fixed line profiling in setup.py

haeb:
with the help of christoph ...

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -38,17 +38,7 @@ test = [
     'tqdm',
     'fire',
     'pycallgraph',  # Used in profiling module
-    (
-        # line_profiler does not work in python 3.7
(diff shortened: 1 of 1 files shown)


boeddeker:
For documentation: `line_profiler` hat a new home under https://github.com/pyutils/line_profiler and the readme starts with 

> NOTICE: This is the official line_profiler repository. The most recent version of line-profiler on pypi points to this repo. The the original line_profiler package by @rkern is currently unmaintained. This fork seeks to simply maintain the original code so it continues to work in new versions of Python.

hence, their intention is to fix our problem that we fixed in the past with a hack and this hack is no longer necessary.
The hack is now removed, because it causes some issues on Windows.

PR #58, opened 2021-10-14
Workaround to enable install into user site directory

thequilo:


TCord:
For documentation purposes: 
The problem is described in  https://github.com/pypa/pip/issues/7953

PR #62, opened 2021-11-10
Windows tests

alexanderwerning:
Adds Windows test environment and modifies or skips tests if necessary.
The most important differences between the systems are:
Tests now allow for WindowsPath objects or backward instead of only forward slashes.
The default numpy integer type is np.int32 on Windows with 64bit Python, instead of np.int64.
Atomic writes are not possible using the os module and modifiying an opened file results in a PermissionError.
Creating processes with the subprocess module produces different outputs since a Windows newline is '\r\n', not '\n' and commands like 'echo' are not found with 'shell=False'.

### DIFF
--- a/paderbox/io/atomic.py
+++ b/paderbox/io/atomic.py
@@ -37,6 +37,11 @@ def open_atomic(file, mode, *args, force=False, **kwargs):
 
     Examples:
 
+    >>> import sys, pytest
+    >>> if sys.platform.startswith('win'):
+    ...     pytest.skip('this doctest does not work on Windows, '
+    ...                 'accessing an opened file is not possible')
+
     Procure a file with some content
     >>> from paderbox.io.cache_dir import get_cache_dir
     >>> file = get_cache_dir() / 'tmp.io.txt'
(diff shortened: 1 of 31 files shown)


paderbox/io/audioread.py
```diff
@@ -202,9 +202,9 @@ def load_audio(
             signal, sample_rate = data, f.samplerate
     except RuntimeError as e:
         if isinstance(path, (Path, str)):
-            from paderbox.utils.process_caller import run_process
-            cp = run_process(['file', f'{path}'])
-            stdout = cp.stdout
+            import magic
+            # recreate the stdout of the 'file' tool
+            stdout = Path(path).as_posix() + ": " + magic.from_file(str(path)) + "\n"
```
boeddeker, line 207: Can you rename `stdout` to something like `msg`?
The `"\n"` can be removed.

paderbox/io/hdf5.py
```diff
@@ -8,6 +8,51 @@
 
 __all__ = ['dump_hdf5', 'update_hdf5', 'load_hdf5']
 
+def join_paths(path, name):
```
boeddeker, line 11: Why not using something like `re.sub('/+', '/', name if name.startswith('/') else f'{path}/{name}')`?
That mirrors `os.path.join` for linux.

paderbox/io/wrapper_load.py
```diff
@@ -240,11 +240,11 @@ def load(
 
 
     To load an unsecure, you have to change `unsafe` to `True`
-    >>> load('/path/to/unsecure_file.pkl', unsafe=False)
+    >>> load('/path/to/unsecure_file.pkl', unsafe=False)  # doctest: +ELLIPSIS
     Traceback (most recent call last):
     ...
     AssertionError: You called Loader.__call__ with unsafe=False
-    for the file /path/to/unsecure_file.pkl.
+    ...
```
boeddeker, line 247: Can you keep `for the file ...unsecure_file.pkl.`?

paderbox/utils/pretty.py
```diff
@@ -71,7 +71,7 @@ def pprint(
      - Support multiple objects (Causes bad readable error in original)
 
 
-    >>> pprint([np.array([1]), np.array([1]*100)])
+    >>> pprint([np.array([1]), np.array([1]*100).astype(np.int64)])
```
boeddeker, line 74: Can you change this to `pprint([np.array([1.]), np.array([1.]*100)])`?
It is not important to use integers here, float64 is also fine.

tests/io_tests/test_hdf5.py
```diff
@@ -7,6 +7,7 @@
 
 
 class TestHdf5:
+    @pytest.mark.skipif(os.name == 'nt', reason="Numpy default integer types differ on Windows.")
```
boeddeker, line 10: How many test breaks? 
For `list`, `tuple`, `array` and `large list` you can change the integers to float.
For `heterogenous list` you can change integers to complex, str or None.

If then only `int` and `np int` fail, you could add if statements.

paderbox/io/atomic.py
```diff
@@ -115,31 +120,41 @@ def open_atomic(file, mode, *args, force=False, **kwargs):
...
+                yield tmp_f
+                if not force:
+                    clean = True
+                    cleanup()
+            finally:
+                if force:
+                    clean = True
+                    cleanup()
+    finally:
+        if clean:
```
boeddeker, line 151: Can you use an if (`if os.name == 'nt'`) statement to minimize the system calls? On network file systems, this is unnecessarily expensive.

boeddeker:
Thank you, LGTM.

boeddeker (approved):

jensheit (approved):

PR #64, opened 2021-11-15
Add SparseArray implementation

thequilo:


paderbox/array/sparse.py
```diff
@@ -0,0 +1,701 @@
...
+def zeros(shape):
+    return SparseArray(shape=shape)
+
+
+def _check_shape(shape1, shape2):
+    if shape1 != shape2:
+        raise ValueError(f'Shape mismatch: {shape1} {shape2}')
+
+
+def _clip(target, array, onset):
```
boeddeker, line 44: I found `_clip` confusing (`np.clip`). Is cat a better name?

paderbox/array/sparse.py
```diff
@@ -0,0 +1,701 @@
...
+    >>> a
+    SparseArray(
+      _SparseSegment(onset=5, length=5: [1 1 1 1 1])
+      _SparseSegment(onset=15, length=5: [2 2 2 2 2]),
+      shape=(20,))
+    >>> np.asarray(a)
+    array([0, 0, 0, 0, 0, 1, 1, 1, 1, 1, 0, 0, 0, 0, 0, 2, 2, 2, 2, 2])
+
+    Overlapping intervals are not allowed
+    >>> a[5:20] = 3
```
boeddeker, line 108: Can you change this to `a[9:11] = 3` or `a[14:16] = 3`? (i.e. hit the corner case)

paderbox/array/sparse.py
```diff
@@ -0,0 +1,701 @@
...
+    # Public constructor interface
+    shape: tuple
+
+    # Private constructor arguments: Don't use _segments directly.
+    # This is required for pt.data.batch.example_to_device to work
+    _segments: list = field(default_factory=list)
+
+    # Other instance attributes
+    pad_value = 0.
+    _is_torch = None
```
boeddeker, line 148: In the source code, you assume `_is_torch` is a bool. Can you make it explicit and set the default to `False`?

paderbox/array/sparse.py
```diff
@@ -0,0 +1,701 @@
...
+    dtype = None
+
+    def __post_init__(self):
+        self.shape = _normalize_shape(self.shape)
+        self._update_dtype(self.dtype)
+
+    @classmethod
+    def from_array_and_offset(cls, array, offset, shape):
```
boeddeker, line 156: This is confusing. We should try to avoid the name `offset`, because it is sometimes the start point and sometimes the endpoint. `onset` has a unique definition.

paderbox/array/sparse.py
```diff
@@ -0,0 +1,701 @@
...
+        out = SparseArray(shape=shape)
+        out._add_segment(_SparseSegment(offset, array))
+        return out
+
+    def as_contiguous(self, dtype=None):
+        arr = self._new_full(dtype=dtype)
+        for segment in self._segments:
+            s, array = _clip(arr, segment.array, segment.onset)
```
boeddeker, line 164: I am confused. I thought, this should never happen. Or do numpy and torch differ?

paderbox/array/sparse.py
```diff
@@ -0,0 +1,701 @@
...
+        if dtype is None and self._segments:
+            self.dtype = self._segments[0].array.dtype
+        if self.dtype is not None:
+            self._is_torch = isinstance(self.dtype, torch.dtype)
+            if self._is_torch:
+                self.pad_value = torch.tensor(pad_value, dtype=self.dtype)
+            else:
+                self.pad_value = self.dtype.type(pad_value)
+        if self._segments:
+            assert all(s.array.dtype == self.dtype for s in self._segments), (
```
boeddeker, line 179: Normally `all` should check all values and return True, hence a list comprehension should be faster. Can you change the generator comprehension to a list comprehension

boeddeker: Another idea, maybe `len(set([...])) == 1` is faster.

paderbox/array/sparse.py
```diff
@@ -0,0 +1,701 @@
...
+        if self.dtype is None:
+            # This is the first segment we add. It defines dtype and pad value
+            self._update_dtype(segment.array.dtype)
+        elif self.dtype != segment.array.dtype:
+            raise TypeError(
+                f'Type mismatch: SparseArray has dtype {self.dtype}, but '
+                f'assigned array has dtype {segment.array.dtype}'
+            )
+
+        if segment.array.shape[:-1] != self.shape[:-1]:
```
boeddeker, line 223: Can you add a hint, that the leading dimensions must be equal? Add least as comment in the source code.

paderbox/array/sparse.py
```diff
@@ -0,0 +1,701 @@
...
+            segment.onset
+        )
+
+        # Check for overlaps with neighboring segments
+        if (
+                len(self._segments) > 0
+                and position > 0
+                and self._segments[position - 1].offset > segment.onset
+        ):
+            import textwrap
```
boeddeker, line 246: Can you move the import to the top of the file?

paderbox/array/sparse.py
TCord: I think you mean `_cut`, but I agree it would be a better fit than `_clip`.

paderbox/array/sparse.py
```diff
@@ -0,0 +1,701 @@
...
+                f'and newly added'
+                f'{textwrap.indent(repr(segment), "  ")}'
+            )
+
+        self._segments.insert(position, segment)
+
+    def __copy__(self):
+        arr = self.__class__(shape=self.shape)
+        for segment in self._segments:
+            arr._add_segment(_SparseSegment(segment.onset, segment.array))
```
boeddeker, line 270: Expensive?

paderbox/array/sparse.py
```diff
@@ -0,0 +1,701 @@
...
+        >>> a[:100] = np.ones(100)
+        Traceback (most recent call last):
+         ...
+        ValueError: Could not assign array with shape (100,) to SparseArray slice with size 10
+        """
+        if isinstance(item, tuple):
+            selector = item[:len(self.shape) - 1]
+            if len(selector) != 1 or selector[0] is not ...:
+                raise NotImplementedError(
+                    'Only ellipsis is allowed for first dimensions in setitem'
```
boeddeker, line 375: I don't get it. Do you want to limit `setitem` to be `[..., slice]` or `[slice]` (i.e. at most 2 arguments and if 2, the first must be Ellipsis?)?

paderbox/array/sparse.py
```diff
@@ -0,0 +1,701 @@
...
+                f'for indexing, not {item!r}'
+            )
+        start, stop = cy_parse_item(item, self.shape)
+        length = stop - start
+
+        if np.isscalar(value):
+            value = self._new_full(shape=self.shape[:-1] + (length,), fill_value=value)
+
+        if isinstance(value, np.ndarray) or isinstance(value, torch.Tensor):
+            if value.shape[-1] != length:
```
boeddeker, line 391: Does `_add_segment` needs checks after this function?

paderbox/array/sparse.py
```diff
@@ -0,0 +1,701 @@
...
+            raise NotImplementedError()
+
+    def __getitem__(self, item):
+        """
+        >>> a = SparseArray(20)
+        >>> a[:10] = 1
+        >>> a[15:] = np.arange(5) + 1
+
+        # Integer getitem
+        >>> a[0], a[10], a[11], a[14], a[15], a[18], a[-1]
```
boeddeker, line 408: `a[10], a[11]` -> `a[9], a[10]`

paderbox/array/sparse.py
```diff
@@ -0,0 +1,701 @@
...
+          _SparseSegment(onset=7, length=3: [1 1 1]),
+          shape=(10,))
+        >>> a[5]
+        Traceback (most recent call last):
+         ...
+        IndexError: index 5 is out of bounds for axis 0 with size 2
+        >>> np.asarray(a[1, 3:5])
+        array([8, 9])
+        """
+        if not isinstance(item, tuple):
```
boeddeker, line 448: Could item be a list?

paderbox/array/sparse.py
```diff
@@ -0,0 +1,701 @@
...
+            item = (item,)
+
+        selector = item[:len(self.shape) - 1]
+        if ... in selector:
+            selector = selector + (slice(None),)
+        if len(item) < len(self.shape):
+            item = slice(None)
```
boeddeker, line 455: Here is an error:
```python
if ... in selector or len(item) == len(self.shape):
    item = item[-1]
elif len(item) < len(self.shape)
    item = slice(None)
else:
    raise ...
```

paderbox/array/sparse.py
```diff
@@ -0,0 +1,701 @@
...
+                index = index + self.shape[-1]
+            if index < 0 or self.shape is not None and index > self.shape[-1]:
+                raise IndexError(
+                    f'Index {item} is out of bounds for ArrayInterval with '
+                    f'shape {self.shape}'
+                )
+            # Could be optimized
+            position = bisect.bisect_right([s.onset for s in self._segments], index)
+            if position == 0:
+                return self.pad_value
```
boeddeker, line 471: When does this happen?

paderbox/array/sparse.py
```diff
@@ -0,0 +1,701 @@
...
+          shape=(10,))
+        >>> c += b
+        >>> c
+        array([ 0,  1,  2,  3,  4,  5,  6,  9, 10, 11])
+        >>> a += c
+        Traceback (most recent call last):
+         ...
+        TypeError: <class 'numpy.ndarray'>
+        """
+        if isinstance(other, _SparseSegment):
```
boeddeker, line 546: Can this happen?

paderbox/array/sparse.py
```diff
@@ -0,0 +1,701 @@
...
+        ValueError: Shape mismatch: (10,) (5,)
+        """
+        if isinstance(other, SparseArray):
+            _check_shape(self.shape, other.shape)
+            self_copy = copy.copy(self)
+            self_copy += other
+            return self_copy
+        else:
+            # Let numpy handle this case
+            return other + self
```
boeddeker, line 526: `return NotImplemented`

paderbox/array/sparse.py
```diff
@@ -0,0 +1,842 @@
...
+        ValueError: Shape mismatch: SparseArray has shape (10,), but assigned array has shape (1, 5).
+        >>> a[:100] = np.ones(100)
+        Traceback (most recent call last):
+         ...
+        ValueError: Could not assign array with shape (100,) to SparseArray slice with size 10
+        """
+        if isinstance(item, tuple):
+            selector = item[:len(self.shape) - 1]
+            # Complex indexing is not supported in __setitem__
+            if len(selector) != 1 or selector[0] is not ...:
```
boeddeker, line 377: What do you want to check here? In `[x, y]` you can use every combination, but in `[x, y, z]`, `x` have to be `...`

paderbox/array/sparse.py
```diff
@@ -0,0 +1,842 @@
...
+                (... in item and item[-1] is ...)
+                or
+                # Ellipsis is not present and no index is present for the last
+                # dimension or the index is slice(None) (select everything)
+                (... not in item and (len(item) < len(self.shape) or item[-1] == slice(None)))
+        ):
+            # The sparse dimension is not indexed, so we can simply forward to
+            # numpy/torch
+            arr = self.__class__(
+                shape=self._segments[0].array[item].shape[:-1] + (self.shape[-1],),
```
boeddeker, line 474: Fails if `self._segments` is empty

TCord (approved):

paderbox/array/sparse.py
```diff
@@ -0,0 +1,1082 @@
...
+            # Infer the dtype from value if the SparseArray doesn't have a
+            # dtype yet
+            value = self._new_full(
+                shape=self.shape[:-1] + (length,), fill_value=value,
+                dtype=self.dtype if self.dtype else _dtype_from_value(value)
+            )
+
+        if isinstance(value, np.ndarray) or isinstance(value, torch.Tensor):
+            # The last dimension has to match the length defined by the indexing
+            if value.shape[-1] != length:
```
boeddeker, line 637: Add an ndim check.

paderbox/array/sparse.py
```diff
@@ -0,0 +1,1082 @@
...
+        >>> zeros((10, 20))[0, :]
+        SparseArray(shape=(20,))
+        >>> zeros((10, 20))[1, 2, 3]
+        Traceback (most recent call last):
+          ...
+        IndexError: too many indices for array: array is 2-dimensional, but 3 were indexed
+        """
+        item = _normalize_item(item, self.shape)
+
+        assert len(item) <= len(self.shape), (item, self.shape)
```
boeddeker, line 713: `assert len(item) == len(self.shape), (item, self.shape)`

boeddeker (approved):

paderbox/array/sparse.py
thequilo: Is implicitly done in the if statement below

boeddeker: ok

PR #65, opened 2021-11-30
Remove torch dependency from SparseArray

thequilo:


boeddeker (approved):
Changes look reasonable, not sure if all dependencies to torch are removed.

PR #66, opened 2021-12-06
update play code

boeddeker:
- Sync docstring format
 - Add class to change defaults, e.g. `play = Play(sample_rate=8000)`
 - Option to return displayable object instead of calling display
 - Ignore sample_rate argument for files

### DIFF
--- a/paderbox/io/play.py
+++ b/paderbox/io/play.py
@@ -50,24 +50,27 @@ def play(
         display: When True, display the audio, otherwise return the widget.
     Returns:
 
+    >>> play(np.array([1, 2, 3]), display=False)
+    <IPython.lib.display.Audio object>
+    >>> play({'a': np.array([1, 2, 3]), 'b': np.array([1, 2, 3])}, display=False)
+    {'a': <play.NamedAudio object>, 'b': <play.NamedAudio object>}
+
     """
     if isinstance(data, dict):
         assert name is None, name
-        for k, v in data.items():
-            play(
-                data=v,
-                name=k,
-
-                channel=channel,
-                sample_rate=sample_rate,
-                size=size,
-                shift=shift,
-                window=window,
-                window_length=window_length,
-                scale=scale,
-                stereo=stereo,
-            )
-        return
+        kwargs = {
+            k: v
+            for k, v in locals().items()
+            if k not in ['data', 'name']
+        }
+        audio = {
+            k: play(data=v, name='.'.join(k), **kwargs )
+            for k, v in data.items()
+        }
+        if display:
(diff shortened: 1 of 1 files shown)


paderbox/io/play.py
```diff
@@ -147,9 +153,78 @@ def _repr_html_(self):
...
+                `channel` parameter and the next channel at the same time.
+            normalize: It true, normalize the data to have values in the range
+                from 0 to 1. Can only be disabled with newer IPython versions.
+
+        Returns:
+
+        """
+        self.kwargs = {
+            **kwargs,
+        }
```
thequilo, line 193: Why copy kwargs?

paderbox/io/play.py
```diff
@@ -147,9 +153,78 @@ def _repr_html_(self):
...
+        audio.name = name
+
+    if display:
+        IPython.display.display(audio)
+    else:
+        return audio
+
+
+class Play:
+    def __init__(self, sample_rate=None, **kwargs):
```
thequilo, line 166: Why not copy the signature over here? That would make the class easier to use (autocomplete)

paderbox/io/play.py
```diff
@@ -20,29 +21,35 @@ def play(
...
+            displayed
+        stereo: If set to true, you can listen to channel as defined by
+            `channel` parameter and the next channel at the same time.
+        normalize: It true, normalize the data to have values in the range
+            from 0 to 1. Can only be disabled with newer IPython versions.
+        display: When True, display the audio, otherwise return the widget.
+    Returns:
+
     """
     if isinstance(data, dict):
```
thequilo, line 54: Just as a suggestion: Support nested dicts and lists, e.g., with `FlatView`? That way one could pass an entire `example['audio_data']` dict to the function and get a nice view in the notebook

jensheit (approved):

paderbox/io/play.py
boeddeker: Good point, in the initial version `sample_rate` had a default argument and was added here.

paderbox/io/play.py
boeddeker: Done

paderbox/io/play.py
boeddeker: While I like the idea to have a nice view, it won't create one for `example['audio_data']`, it would simply pollute the notebook (i.e. one play for each channel for each audio file, which may break for long source files).
I think in most cases it is better to call it manually, this function has already more magic as I like.

michael-kuhlmann (approved):

PR #69, opened 2022-02-24
Add workflow for Windows wheels

alexanderwerning:
Adds github workflow generating wheels for Windows.
The pipeline can be triggered via the github UI ('workflow_dispatch') or is automatically started on version tags or branches.
Wheels are created for 32 and 64 bit Windows and for Python 3.6, 3.7, 3.8 and 3.9

For wheels created by the workflow see: https://github.com/alexanderwerning/paderbox/actions/runs/1887543159

### DIFF
--- a/.github/workflows/wheels_windows.yml
+++ b/.github/workflows/wheels_windows.yml
@@ -1,0 +1,49 @@
+name: Build CI wheels for windows
+
+on:
+  # Manual trigger
+  workflow_dispatch:
+  # Save CI by only running this on release branches or tags.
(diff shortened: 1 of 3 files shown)


boeddeker:
Thank you. LGTM. 

Do you know if the wheels can be download via terminal? i.e. it would be nice, when the [maintenance.md](https://github.com/fgnt/paderbox/blob/master/maintenance.md) could be updated, so that also the windows wheels are included.

boeddeker (approved):

PR #74, opened 2022-04-04
Add groupby utility function

thequilo:


paderbox/utils/iterable.py
```diff
@@ -0,0 +1,79 @@
+import collections
+import operator
+
+from typing import Callable, Any, Hashable, Optional, Union, Iterable
+
```
boeddeker, line 5: add an `__all__`. It makes it more obvious, that zip is not public

paderbox/utils/iterable.py
```diff
@@ -0,0 +1,79 @@
...
+        key_iterable = iterable
+    elif isinstance(key, (str, int)):
+        key_iterable = map(operator.itemgetter(key), iterable)
+    else:
+        key_iterable = key
+
+    groups = collections.defaultdict(list)
+    for key, value in zip(key_iterable, iterable, strict=True):
+        groups[key].append(value)
+    return groups
```
boeddeker, line 79: `return dict(groups)`

paderbox/utils/iterable.py
```diff
@@ -0,0 +1,92 @@
...
+        >>> groupby('abc', ['a'])
+        Traceback (most recent call last):
+            ...
+        ValueError: zip() argument 2 is longer than argument 1
+    """
+    value_getter = identity
+    if callable(key):
+        key_fn = key
+    elif key is None:
+        key_fn = identity
```
boeddeker, line 81: Why not None?

paderbox/utils/iterable.py
```diff
@@ -0,0 +1,92 @@
...
+    elif isinstance(key, (str, int)):
+        key_fn = operator.itemgetter(key)
+    else:
+        key_fn = operator.itemgetter(1)
+        value_getter = operator.itemgetter(0)
+        iterable = zip(key, iterable, strict=True)
```
boeddeker, line 87: `iterable = zip(iterable, key, strict=True)` Produces a more natural error message.
Currently, `key_fn` and `value_getter` are swapped

boeddeker (approved):

PR #76, opened 2022-05-25
Replace deprecated pprint_val in context_manager.py

tobvogel:
In order to fix an error in the LaTeX-Export notebook, i got the error that the method pprint_val was not found. It seems to be no longer present in ScalarFormatter class of newer matplotlib versions and could possibly be replaced with format_data_short to achieve the result.


jensheit (approved):
pprint_value was deppricated in matplotlib 3.1

PR #85, opened 2023-03-14
Allow slicing notation in audio file names

thequilo:
Add notation to select channels and a slice from the audio path passed to `load_audio`

### DIFF
--- a/paderbox/io/audioread.py
+++ b/paderbox/io/audioread.py
@@ -78,6 +78,9 @@ def load_audio(
     stop : int, optional
         The index after the last frame to be read.  A negative value
         counts from the end.  Not allowed if `frames` is given.
+    channel : int or slice or list of int or tuple of int
+        Channel(s) to select from a multichannel audio file. Can be anything
+        that can index a numpy array along a single dimension.
     dtype : {'float64', 'float32', 'int32', 'int16'}, optional
         Data type of the returned array, by default ``'float64'``.
         Floating point audio data is typically in the range from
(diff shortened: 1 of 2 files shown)


paderbox/io/audioread.py
```diff
@@ -25,6 +25,7 @@ def load_audio(
         frames=-1,
         start=0,
         stop=None,
+        channel=None,
```
boeddeker, line 28: Add to docstring

paderbox/io/audioread.py
```diff
@@ -228,17 +269,83 @@ def load_audio(
...
 
-    # When signal is multichannel, than soundfile return (samples, channels)
+    # When signal is multichannel, then soundfile returns (samples, channels)
     # At NT it is more common to have the shape (channels, samples)
     # => transpose
     signal = signal.T
 
+    # Slice along channel dimension if channel_slice is given
+    if channel is not None:
+        signal = signal[channel,]
```
boeddeker, line 279: Check that signal has two axis. Soundfile does a squeeze by default

boeddeker:
We should document, that we used pytest as inspiration for the notation

thequilo:
The tests work locally for me (Python 3.10). I have to check what goes wrong on GitHub

paderbox/io/audioread.py
```diff
@@ -228,17 +261,84 @@ def load_audio(
...
 
-    # When signal is multichannel, than soundfile return (samples, channels)
+    # When signal is multichannel, then soundfile returns (samples, channels)
     # At NT it is more common to have the shape (channels, samples)
     # => transpose
     signal = signal.T
 
+    # Slice along channel dimension if channel_slice is given
+    if channel is not None:
+        assert signal.ndim == 2 or channel == 0, (signal.shape, channel)
```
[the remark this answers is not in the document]
thequilo: Exactly. I think this shouldn't fail when `channel = 0`  and a file with only a single channel. But I realize now that then also `channel = slice(None)` should pass through

paderbox/io/audioread.py
```diff
@@ -121,6 +126,27 @@ def load_audio(
...
+    >>> data = load_audio(path)
+    >>> data.shape
+    (6, 38520)
+    >>> data = load_audio(str(path) + '::[:,0]')
+    >>> data.shape
+    (38520,)
+    >>> data = load_audio(str(path) + '::[:,:2]')
+    >>> data.shape
+    (2, 38520)
+    >>> data = load_audio(str(path) + '::[:,2:4]')
```
boeddeker, line 144: Can you add a test with pathlib.Path?

boeddeker (approved):

PR #86, opened 2023-05-02
Allow to construct a SparseArray from arrays and onsets

thequilo:


paderbox/array/sparse.py
```diff
@@ -397,16 +408,51 @@ def from_array_and_onset(cls, array, onset, shape):
         array([1., 1., 0., 0., 0., 0., 0., 0., 0., 0.])
         >>> SparseArray.from_array_and_onset(np.ones(5), 7, 10)
         SparseArray(_SparseSegment(onset=7, array=array([1., 1., 1.])), shape=(10,))
+        >>> SparseArray.from_array_and_onset(np.ones(5), 7).as_contiguous()
+        array([0., 0., 0., 0., 0., 0., 0., 1., 1., 1., 1., 1.])
         """
+        if shape is None:
+            shape = list(array.shape)
+            shape[0] += onset
```
boeddeker, line 416: change to -1

paderbox/array/sparse.py
```diff
@@ -397,16 +408,51 @@ def from_array_and_onset(cls, array, onset, shape):
...
+            if shape[0] < 0:
+                shape[0] = 0
         out = SparseArray(shape=shape)
-        if onset < 0:
-            array = array[..., -onset:]
-            onset = 0
-        if array.shape[-1] + onset > out.shape[-1]:
-            array = array[..., :out.shape[-1] - onset]
-        out._add_segment(_SparseSegment(onset, array))
+        out._add_segment(_SparseSegment(onset, array).get_clipped(out.shape[-1]))
```
boeddeker, line 420: Should we disable, that _SparseSegment could represent a segment, that is partially outside?

paderbox/array/sparse.py
thequilo: Done

paderbox/array/sparse.py
thequilo: Added a method  `_SparseSegment.from_array` that constructs a clipped segment. This should be used everywhere where a `_SparseSegment` is constructed from an array that is not controlled by a SparseArray

paderbox/array/sparse.py
```diff
@@ -397,16 +397,51 @@ def from_array_and_onset(cls, array, onset, shape):
...
+        >>> SparseArray.from_arrays_and_onsets([np.ones(2), np.ones(1)], [0, 3]).as_contiguous()
+        array([1., 1., 0., 1.])
+        >>> SparseArray.from_arrays_and_onsets([np.ones(2), np.ones(1)], [0, 3], (10,)).as_contiguous()
+        array([1., 1., 0., 1., 0., 0., 0., 0., 0., 0.])
+        """
+        if shape is None:
+            for array, onset in zip(arrays, onsets):
+                if shape is None or array.shape[-1] + onset > shape[-1]:
+                    shape = list(array.shape)
+                    shape[-1] += onset
```
boeddeker, line 424: Suggestion: Add helper function, that predicts the shape.

Why? `from_array_and_onset` has
```
            if shape[-1] < 0:
                shape[-1] = 0
```
while it is missing in `from_arrays_and_onsets`.

paderbox/array/sparse.py
```diff
@@ -397,16 +397,51 @@ def from_array_and_onset(cls, array, onset, shape):
...
+
+        out = SparseArray(shape=shape, _pad_value=_pad_value_like(arrays[0], pad_value))
+
+        assert len(arrays) == len(onsets)
```
boeddeker, line 428: `assert len(arrays) == len(onsets), (len(arrays), len(onsets))`

paderbox/array/sparse.py
```diff
@@ -397,16 +397,51 @@ def from_array_and_onset(cls, array, onset, shape):
...
+
+        for array, onset in zip(arrays, onsets):
+            out._add_segment(_SparseSegment.from_array(array, onset, out.shape[-1]))
+
         return out
 
+    def as_arrays_and_onsets(self):
```
boeddeker, line 435: Add a small sentence with a motivation for this function?

PR #87, opened 2023-06-30
ArrayInterval: slice property, more element wise comparisons and some minor changes

boeddeker:


paderbox/array/interval/core.py
```diff
@@ -824,9 +889,24 @@ def __eq__(self, other):
...
+
+    def __le__(self, other):
+        if not isinstance(other, ArrayInterval): return NotImplemented
+        else: return _combine(operator.__le__, self, other)
+
+    def __gt__(self, other):
+        if not isinstance(other, ArrayInterval): return NotImplemented
+        else: return _combine(operator.__gt__, self, other)
+
+    def __ge__(self, other):
```
thequilo, line 906: Can you add `__ne__`?

paderbox/array/interval/core.py
```diff
@@ -701,6 +764,9 @@ def sum(self, axis=None, out=None):
             sum = self.shape[0] - sum
         return sum
 
+    def mean(self, axis=None, out=None):
+        return self.sum(axis, out) / self.shape[0]
```
thequilo, line 768: This fails when someone adds support for `out` to `self.sum`. I guess it's fine here, but we should keep that in mind.

paderbox/array/interval/core.py
```diff
@@ -677,6 +680,66 @@ def __getitem__(self, item):
...
+
+            shape, = [None] if self.ai.shape is None else self.ai.shape
+
+            if stop is None: stop = shape
+            if start is None: start = 0
+
+            if shape is None:
+                assert start >= 0, (item, start, stop)
+                assert stop >= 0, (item, start, stop)
+                if stop is None:
```
thequilo, line 726: `stop` can't be `None` here, it would fail in  the `assert` before

paderbox/array/interval/core.py
```diff
@@ -677,6 +680,66 @@ def __getitem__(self, item):
...
+            self.ai = ai
+
+        def __getitem__(self, item):
+            start, stop = cy_parse_item(item, self.ai.shape)
+            intervals = cy_intersection((start, stop), self.ai.normalized_intervals)
+
+            shape, = [None] if self.ai.shape is None else self.ai.shape
+
+            if stop is None: stop = shape
+            if start is None: start = 0
```
thequilo, line 721: `start` and `stop` can't be `None` because `cy_parse_interval` always returns integers or raises an exception

paderbox/array/interval/core.py
```diff
@@ -677,6 +680,66 @@ def __getitem__(self, item):
...
+                assert stop >= 0, (item, start, stop)
+                if stop is None:
+                    shape = None
+                else:
+                    shape = stop - start
+                    assert shape >= 0, (shape, item, start, stop)
+            else:
+                # Convert negative index to positive index
+                start %= shape
+                stop %= shape
```
thequilo, line 734: `cy_parse_item` already converts negative indices up to `-shape[0]`. Just check that they are not negative

paderbox/array/interval/core.py
boeddeker: Done

paderbox/array/interval/core.py
boeddeker: Fixed

paderbox/array/interval/core.py
boeddeker: Fixed. Now shape=None is supported

paderbox/array/interval/core.py
boeddeker: Removed. All checks and modifications are done in `cy_parse_item`

paderbox/array/interval/core.py
boeddeker: Fixed

thequilo (approved):

PR #89, opened 2023-07-18
Improve pprint

boeddeker:


paderbox/utils/pretty.py
```diff
@@ -53,6 +53,50 @@ def _ipy_pprint_tensor(obj, p, cycle):
         self.type_pprinters[type({}.items())] = \
             IPython.lib.pretty._seq_pprinter_factory('dict_items(', ')')
 
+    @property
+    def max_seq_length(self):
+        if isinstance(self._max_seq_length, int):
+            return self._max_seq_length
+        elif isinstance(self._max_seq_length, (tuple, list)):
```
thequilo, line 60: the check for tuple/list is enough here, the check for int is redundant since `self._max_seq_length` is returned at the bottom

boeddeker: Good point. Removed

thequilo (approved):

PR #90, opened 2023-07-20
Fix function cos_similarity

RautenbergFrederik:
If dimension of input is (1,N) or (N,1), cos_similarity was not calculated.

### DIFF
--- a/paderbox/math/vector.py
+++ b/paderbox/math/vector.py
@@ -25,7 +25,10 @@ def cos_distance(a, b):
     :param b: vector b (1xN or Nx1 numpy array)
     :return: distance (scalar)
     """
-    return 0.5 * (1 - sum(a * b) / np.sqrt(sum(a ** 2) * sum(b ** 2)))
+    assert a.shape == b.shape, 'Both vectors must have the same dimension'
+    assert a.squeeze().ndim == 1 and b.squeeze().ndim == 1, \
+        f"Input must be vectors: {a.shape} {b.shape}"
+    return 0.5 * (1 - np.sum(a * b) / np.sqrt(np.sum(a ** 2) * np.sum(b ** 2)))
 
 
 def normalize_vector_to_unit_length(data):


boeddeker:
Could you add an axis argument? I think this is one of the functions that we translated from MATLAB and never changed to numpy style.

paderbox/math/vector.py
```diff
@@ -25,7 +25,10 @@ def cos_distance(a, b):
     :param b: vector b (1xN or Nx1 numpy array)
     :return: distance (scalar)
     """
-    return 0.5 * (1 - sum(a * b) / np.sqrt(sum(a ** 2) * sum(b ** 2)))
+    assert a.shape == b.shape, 'Both vectors must have the same dimension'
```
boeddeker, line 28: `assert a.shape == b.shape, f'Both vectors must have the same dimension. {a.shape} {b.shape}'`

RautenbergFrederik:
In the cos similarity function, the absolute value is used in the denominator. So for the case [1,0] [-1,0] the similarity would not be -1 but 1, which conflicts with the definition of cos similarity?

boeddeker:
There are different definitions for the cosine similarity and the cosine distance.
Depending, what you do, you use a different definition.
When we wrote those two function, we discussed, which implementation we should use and 
for some reason, we decided to use the current split. (I forgot the arguments)

When you search for the definition of cosine similarity, you find the definition for real valued numbers without conj and without abs inside, so the old implementation of "cos_distance" was correct, according to this definition.

When you work with complex number, you have to fix the denominator, but the number would still be complex.
As far as I know, the abs is usually used, when you work with complex numbers and cosine dist or similarity.

Since you changed the implementation to use abs, could you give a motivation/use case for abs?

@TCord What do you usually use?

While I am fine with a breaking change to remove support for column vectors (shouldn't cause any trouble), we should be careful with a breaking change that introduces an abs.

TCord:
I would vote against making a breaking change, here. We should keep the behavior of the function (value range between 0 and 1, but no usage of an absolute value) identical, and instead add an axis argument and an assertion that the output will be valid.

PR #92, opened 2023-11-10
add pb.array.kernel

boeddeker:


paderbox/array/kernel.py
```diff
@@ -0,0 +1,208 @@
...
+    a = np.array(a)
+    values = np.array(list(' ▁▂▃▄▅▆▇█'))
+    m = np.amax(a)
+    if m == 0:
+        m = 1
+    a = np.round(a * ((len(values)-1) / m)).astype(int)
+    print(repr(''.join(values[a])))
+
+
+def np_kernel(
```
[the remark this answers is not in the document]
boeddeker: That got lost in my code cleanup. I change it.

paderbox/array/kernel.py
```diff
@@ -0,0 +1,208 @@
...
+):
+    """
+    Apply a kernel to the last axis of x.
+
+    Args:
+        x: np.array
+        kernel_size:
+        kernel: a function that should be applied to the kernel window.
+            Has to accept an axis argument, like np.mean, np.amax, np.median.
+        padding_mode:
```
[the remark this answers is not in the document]
boeddeker: I add som text

paderbox/array/kernel.py
```diff
@@ -0,0 +1,208 @@
...
+    '   ██████   '
+    >>> _plot(a); _plot(np_kernel(a, 3, kernel=np.amin))
+    '    ████    '
+    '     ██     '
+    >>> _plot(a); _plot(np_kernel(a, 3, kernel=np.mean))
+    '    ████    '
+    '   ▃▅██▅▃   '
+
+    """
+    assert kernel_size % 2 == 1, (kernel_size, 'must be odd.')
```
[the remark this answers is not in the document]
boeddeker: done

paderbox/array/kernel.py
```diff
@@ -0,0 +1,208 @@
...
+    y = kernel(pb.array.segment_axis(x, kernel_size, 1, end='pad'), axis=-1)
+
+    if pad_position == 'post':
+        shift = kernel_size // 2
+        padding = [(0, 0)] * (y.ndim - 1) + [(shift, shift)]
+        y = np.pad(y, padding, padding_mode)
+    return y
+
+
+def _ai_dilate_erode(ai, khalf):
```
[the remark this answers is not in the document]
boeddeker: Yes, it is the core of `ai_dilate` and `ai_erode`.

TCord (approved):

paderbox/array/kernel.py
```diff
@@ -0,0 +1,212 @@
...
+    Apply a kernel to the last axis of x.
+
+    Args:
+        x: np.array
+        kernel_size:
+        kernel: a function that should be applied to the kernel window.
+            Has to accept an axis argument, like np.mean, np.amax, np.median.
+        padding_mode:
+            The mode for np.pad, see np.pad for the options.
+        pad_position:
```
thequilo, line 49: I was confused by the parameter name. This sounds like: `pre`: pad beginning, `post`: pad end. I don't have a strong opinion, but maybe just `pad`?

paderbox/array/kernel.py
```diff
@@ -0,0 +1,212 @@
...
+    '   ██████   '
+    >>> _plot(a); _plot(np_kernel1d(a, 3, kernel=np.amin))
+    '    ████    '
+    '     ██     '
+    >>> _plot(a); _plot(np_kernel1d(a, 3, kernel=np.mean))
+    '    ████    '
+    '   ▃▅██▅▃   '
+
+    """
+    assert kernel_size % 2 == 1, (kernel_size, 'kernel size has to be odd.')
```
thequilo, line 68: Is it better to follow Python style guidelines and use `ValueError` for parameter checks? `asserts` can be deactivated by the user and carry less information

paderbox/array/kernel.py
```diff
@@ -0,0 +1,212 @@
...
+
+
+def np_kernel1d(
+        x,
+        kernel_size,  # an odd number
+        *,
+        kernel: callable,  # e.g. np.mean, np.amax, np.amin
+        padding_mode: 'typing.Literal["edge"]' = 'edge',
+        pad_position: 'typing.Literal["pre", "post", None]' = 'pre',
+):
```
thequilo, line 38: Does it make sense to add an `axis` argument? You could use `np.moveaxis` to move it to the back, apply the same code you already have, and then move the axis back to its original position.

paderbox/array/kernel.py
```diff
@@ -0,0 +1,212 @@
...
+    """
+    pairs = np.array(ai.normalized_intervals)
+    assert pairs.shape[-1] == 2, pairs.shape
+    if ai.inverse_mode:
+        khalf = -khalf
+    pairs[:, 0] -= khalf
+    pairs[:, 1] += khalf
+    pairs = np.maximum(pairs, 0)
+    if ai.shape is not None:
+        pairs = np.minimum(pairs, ai.shape[-1])
```
thequilo, line 108: This is already done in `from_pairs`:

```python
In [7]: ArrayInterval.from_pairs([(0,2)], shape=(1,))
Out[7]: ArrayInterval("0:1", shape=(1,))
```

paderbox/array/kernel.py
```diff
@@ -0,0 +1,212 @@
...
+    '██████████'
+
+    """
+    pairs = np.array(ai.normalized_intervals)
+    assert pairs.shape[-1] == 2, pairs.shape
+    if ai.inverse_mode:
+        khalf = -khalf
+    pairs[:, 0] -= khalf
+    pairs[:, 1] += khalf
+    pairs = np.maximum(pairs, 0)
```
thequilo, line 106: Should this be done in `from_pairs`? It currently fails with negative indices and I'm not sure if it makes sense to clip to 0 or if it's better to fail in general.

paderbox/array/kernel.py
```diff
@@ -0,0 +1,212 @@
...
+    In case of a boolean arrays, this operation is known as dilation.
+
+    >>> a = np.array([0, 0, 0, 0, 1, 1, 1, 1, 0, 0, 0, 0,])
+    >>> _plot(a); _plot(max_kernel1d(a, 3))
+    '    ████    '
+    '   ██████   '
+    """
+    if isinstance(x, np.ndarray):
+        return np_kernel1d(x, kernel_size, kernel=np.amax)
+    elif hasattr(x, 'normalized_intervals'):
```
thequilo, line 161: Why no `isinstance` check?

paderbox/array/kernel.py
TCord: In the other paderbox functions, `pad` always is a boolean parameter that defines whether there ist padding. So, I would not use that name here.

paderbox/array/kernel.py
TCord: The axis argument would not have an effect if ArrayIntervals are used, right? I thunk that's not a behavior we want.

paderbox/array/kernel.py
boeddeker: Should it then be something like `edge in ['pre_pad', 'post_pad', 'cut']`?

I think `pad` is too general. IMHO the `position` triggers the user to check the documentation.
Or is there a better name for the execution order?

At the moment, I have a favor for keeping the name.

paderbox/array/kernel.py
boeddeker: The way you formulate it, the "assert" is correct here.
You let the program run once and see, if it doesn't fail.
Then you disable all asserts and get rid of the overhead for the checks.

paderbox/array/kernel.py
boeddeker: This shouldn't be done in `from_pairs`. Negative indices in python count from the end.
`from_pairs` can be interpreted as an alternative of
```python
ai = zeros(shape=...)
for s,e in pairs:
    ai[s:e] = True
```

We may want to change `from_pairs` to fail for negative indices
```python
>>> ArrayInterval_from_pairs([[-4, 2]], shape=5)
ArrayInterval("1:2", shape=(5,))
```
But that is an independent discussion.

paderbox/array/kernel.py
boeddeker: I'd like to provide the correct values to `from_pairs`. That makes the code more robust to future changes (e.g. new datatypes)

paderbox/array/kernel.py
thequilo: I didn't find a better alternative for position, unfortunately. I don't like `edge` either. Then keep the name

paderbox/array/kernel.py
boeddeker: I don't want to trigger the import. ArrayInterval is cython.

paderbox/array/kernel.py
thequilo: This is a function that is used outside of `paderbox`. In that case, `ValueError` is preferred. But that's also a matter of taste

paderbox/array/kernel.py
thequilo: `ArrayInterval`s are always 1D, so forcing `axis=0` is reasonable. This function could also be useful outside of this module where `ArrayInterval`s are not needed.

boeddeker: Done. `np_kernel1d` has now an axis argument, but I kept the thin wrappers with the minimum number of arguments, so they can work with different backends.

paderbox/array/kernel.py
boeddeker: I don't care about the Exception type, when the exception should never be caught.
Hyperparameters are such a case for me. I find asserts are easier to read and more compact.
It is ok to deactivate `asserts`. The idea is to reduce the overhead, and that is the case here.

paderbox/array/kernel.py
boeddeker: We can add the axis argument to the wrappers, once we have a use case.

thequilo (approved):

PR #94, opened 2023-11-27
plot: int as color, visible_dB and x/y lim support for images

boeddeker:


paderbox/visualization/plot.py
```diff
@@ -48,6 +48,51 @@ def wrapper(*args, **kwargs):
     return wrapper
 
 
+def allow_int_color(f):
```
thequilo, line 51: What is the difference to `f'C{number}'`?

boeddeker: Yes, I also prefer native solutions, but I didn't know `f'C{number}'`.
I would either remove the code or raise a better exception message.
Matplotlibs exception message is not helpful: `1 is not a valid value for color` (No hint, that the user most likely wanted to find `f'C{number}'`. At least that is what I did several years ago.).

boeddeker: Done

thequilo (approved):

PR #96, opened 2023-12-21
some updates to module_stft , module_resample, random_utils and io.download

JanekEbb:


paderbox/utils/random_utils.py
```diff
@@ -177,6 +179,16 @@ def _sample(self, shape):
         return truncexpon(self.truncation / self.scale, self.loc, self.scale).rvs(shape)
 
 
+@dataclasses.dataclass
+class Choice(_Sampler):
+    events: int = 1
```
boeddeker, line 184: `Choice` and `choice` should have the same default for `events`.
While other distributions have some kind of natural defaults, I would say, we shouldn't set a default for events.

paderbox/io/download.py
```diff
@@ -1,184 +1,142 @@
...
-            os.remove(local_file)
-
-
-def download_file_list(file_list, target_dir, exist_ok=False, logger=None):
+    assert local_file.exists(), local_file
+    if target_dir is None:
+        target_dir = local_file.parent
+    else:
+        target_dir = Path(target_dir)
+    if not target_dir.exists():
```
boeddeker, line 61: Remove this if, just call `target_dir.mkdir(parents=True, exist_ok=True)`

paderbox/io/download.py
```diff
@@ -1,184 +1,142 @@
...
+                        # Delete latest file, since most likely it
+                        # was not extracted fully
+                        if target_file.exists():
+                            os.remove(target_file)
+                        raise
+                elif not exist_ok:
+                    raise FileExistsError(target_file)
+                tar.members = []
+        os.remove(local_file)
+
```
boeddeker, line 100: Is there a reason, that you didn't define an else path with a raise? If not, could you raise a warning, when there is an unsupported suffix?

paderbox/io/download.py
```diff
@@ -1,184 +1,142 @@
...
+    if isinstance(extract, bool):
+        extract = len(file_list) * [extract]
+    assert len(extract) == len(file_list), (len(extract), len(file_list))
+    if num_workers > 1:
+        with ProcessPoolExecutor(num_workers) as ex:
+            for _ in ex.map(
+                download_file,
+                file_list,
+                [target_dir / Path(f).name.split('?')[0] for f in file_list],
```
boeddeker, line 129: Could you add a comment, why only name is used and why everything behind the `?` is ignored?
Alternatively, a doctest could also help to understand the motivation.

paderbox/transform/module_stft.py
```diff
@@ -344,6 +347,8 @@ def sample_index_to_stft_frame_index(sample, window_length, shift, fading='full'
 
     >>> [sample_index_to_stft_frame_index(i, 8, 1, fading=None) for i in range(12)]
     [0, 0, 0, 0, 1, 2, 3, 4, 5, 6, 7, 8]
+    >>> sample_index_to_stft_frame_index(np.arange(12), 8, 1, fading=None)
```
boeddeker, line 350: Add a dtype for arange. The Windows test fails, because the default is int32 on windows.

paderbox/io/download.py
```diff
@@ -1,184 +1,143 @@
...
+def download_file(remote_file, local_file, exist_ok=False, extract=False):
     """
     Download single file to local_dir
 
     Args:
         remote_file:
         local_file:
         exist_ok:
+        extract:
+        progress_par:
```
thequilo, line 22: `progress_bar` is no arg

paderbox/utils/random_utils.py
```diff
@@ -177,6 +179,21 @@ def _sample(self, shape):
         return truncexpon(self.truncation / self.scale, self.loc, self.scale).rvs(shape)
 
 
+@dataclasses.dataclass
+class Choice(_Sampler):
+    events: int = None
```
boeddeker, line 184: Why not simply `events: int`? (without a default value)
Then the `__post_init__``wouldn't be nessesary.

boeddeker (approved):
When removing the default for events, LGTM.

PR #97, opened 2024-02-28
Update pandas_utils.py for column names with line breaks

mdeegen:
in is_valid_variable_name because of exception when column name included a new line


boeddeker:
Can you extend the doctest for non-valid variable names to reflect test this bugfix?

PR #98, opened 2024-03-15
Update core.py

jkienegger:
make inverse_mode to keyword only

### DIFF
--- a/paderbox/array/interval/core.py
+++ b/paderbox/array/interval/core.py
@@ -195,7 +195,7 @@ class ArrayInterval:
     from_str = staticmethod(ArrayInterval_from_str)
     inverse_mode = False
 
-    def __init__(self, array, *ignore, inverse_mode: bool = False):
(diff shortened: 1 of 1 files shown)


paderbox/array/interval/core.py
```diff
@@ -195,7 +195,7 @@ class ArrayInterval:
     from_str = staticmethod(ArrayInterval_from_str)
     inverse_mode = False
 
-    def __init__(self, array, inverse_mode: bool = False):
+    def __init__(self, array, *ignore, inverse_mode: bool = False):
```
thequilo, line 198: `def __init__(self, array, *, inverse_mode: bool = False):` does exactly what you want without the `TypeError` below.

jkienegger: So we know, e.g. the parameter we gave has no purpose?

jkienegger: Ok, I understand, i didn't know it worked like this

PR #101, opened 2024-04-26
fixed deprecated bug python 3.11 - np.int to int

schmalen:
Proposal to fix deprecated code to run on python 3.11


thequilo (approved):

boeddeker (approved):

boeddeker:
For documentation: This is a numpy version issue. Numpy removed `np.int`, which was an alias of `int`.

